PR #8, opened 2019-03-06
update request code because it's greater than 16 bits

Dviejopomata:
I had a problem when extending from FlutterFragmentActivity instead of FlutterActivity and it's because the requestCode in **startActivityForResult** is greater than 16 bits ( 65535) i changed it to 65030 and 65031.


MaikuB:
Thanks!

PR #41, opened 2019-09-09
Small update to the Readme file.

Omortis:
To get around the Chrome Custom Tabs issue you need to add AndroidX support to the app. I think the existing paragraph is out of date (Android Studio is at 3.5 now).


MaikuB:
If you look at the link in the note, the details are still relevant. Whilst Android Studio is at 3.5 now, developers upgrade at their pace so the note is there for those that encounter the issue. Furthermore, the plugin has always required AndroidX but there was a bug specific to Android in the tooling in the past that I have linked to. As such I won't be taking this PR but will look at stating that the plugin requires AndroidX more clearly. Just a sidenote though, I believe adding the two lines you've written to `gradle.properties` isn't sufficient as the dependencies need to be updated too

Omortis:
Yes, I kind of meant "add these 2 lines" in addition to your directions. Those two lines exist in `gradle.properties` in this repo but are _not_ added to a newly created flutter application by default, and that tripped me up for a bit. The lines added were more of a suggestion ;-)

Issue #12, opened 2019-03-29
how can we use http as well or https with self signed certificate 


harpreetfervaha:
In the android we can use  ConnectionBuilder}that permits connecting to http links, and ignores certificates for https connections. 
is there any way to use http instead of https.
I am getting this error if I am trying to use http

E/AndroidRuntime( 4341): java.lang.RuntimeException: An error occurred while executing doInBackground()
E/AndroidRuntime( 4341): 	at android.os.AsyncTask$3.done(AsyncTask.java:353)
E/AndroidRuntime( 4341): 	at java.util.concurrent.FutureTask.finishCompletion(FutureTask.java:383)
E/AndroidRuntime( 4341): 	at java.util.concurrent.FutureTask.setException(FutureTask.java:252)
E/AndroidRuntime( 4341): 	at java.util.concurrent.FutureTask.run(FutureTask.java:271)
E/AndroidRuntime( 4341): 	at android.os.AsyncTask$SerialExecutor$1.run(AsyncTask.java:245)
E/AndroidRuntime( 4341): 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1162)
E/AndroidRuntime( 4341): 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:636)
E/AndroidRuntime( 4341): 	at java.lang.Thread.run(Thread.java:764)
E/AndroidRuntime( 4341): Caused by: java.lang.IllegalArgumentException: only https connections are permitted
 And can we use https with self signed certificate

MaikuB:
Not at this stage, given you seem to be familiar with what's required and require this functionality,  are you able to submit a PR?

harpreetfervaha:
so its don't support https with a self-signed certificate?

MaikuB:
My understanding is you need make use of the connection builder for either http or https with self-signed certificates and your original post has said as much so as per my previous reply, not at this stage. The code is there for you to confirm and you should be able to verify this if needed. The identity providers I have tested this with have also been mentioned in the readme

balu-ce:
@harpreetfervaha  Can you able to solve this issue. I cannot able to use simply http.

harpreetfervaha:
> @harpreetfervaha Can you able to solve this issue. I cannot able to use simply http.

[...]

PR #60, opened 2019-11-21
InsecureConnections for Android

Cir0X:
Since I had some [problems](https://github.com/MaikuB/flutter_appauth/issues/12#issuecomment-554936007) while using unencrypted http connections for development purposes, I've found the solution:
The `AuthorizationServiceConfiguration.fetchFromUrl(...)` call need to receive a `ConnectionBuilder` as [third](https://github.com/openid/AppAuth-Android/blob/f1dbeb51f4d905523d804d70363abdb00728d99c/library/java/net/openid/appauth/AuthorizationServiceConfiguration.java#L246) argument.

Furthermore it is necessary to make some changes to your app, in order to allow unencrypted http traffic:
* For [Android](https://stackoverflow.com/a/51902630/3014979) add `android:usesCleartextTraffic="true"` to the `<application` tag in the `AndroidManifest.xml` 
* For [iOS](https://stackoverflow.com/a/40299837/3014979) add the following to the `Info.plist`:
```xml
<key>NSAppTransportSecurity</key>
<dict>
    <key>NSAllowsArbitraryLoads</key>
    <true/>
</dict>
```

### DIFF
--- a/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
+++ b/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
@@ -153,7 +153,7 @@ public class FlutterAppauthPlugin implements MethodCallHandler, PluginRegistry.A
                 }
             };
             if (tokenRequestParameters.discoveryUrl != null) {
-                AuthorizationServiceConfiguration.fetchFromUrl(Uri.parse(tokenRequestParameters.discoveryUrl), callback);
+                fetchFromUrl(Uri.parse(tokenRequestParameters.discoveryUrl), callback);
             } else {
                 AuthorizationServiceConfiguration.fetchFromIssuer(Uri.parse(tokenRequestParameters.issuer), callback);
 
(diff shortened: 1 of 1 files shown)


MaikuB:
Do you happen to have an example app that makes use of these changes and also points to a demo server so I could try out the changes?

Cir0X:
Sorry I just have local development server, which is not public...

MaikuB:
It's possible to setup ngrok to provide a public tunnel but I can understand if you're hesitant to set that up. What kind of server did you setup and are you able to provide details how I could do (e.g. tutorials)

Cir0X:
The server isn't written by myself, but it is build with the [Identity Server](https://identityserver.io/) .NET Framework. 
I check if I can build a simple demo app and provide an ngrok tunnel for you.

MaikuB:
Any updates? One thing I noticed is your changes is that `InsecureConnectionBuilder.INSTANCE` is passed to `fetchFromUrl`. There isn't the equivalent for `fetchFromIssuer`.

MaikuB:
Closing since I don't have a way to evaluate this and I don't think it's enough to fix the issue based on my previous

PR #80, opened 2020-02-08
Updated Android to send app auth request on a separate thread to prevent blocking the UI thread in Flutter

acoutts:
The library currently blocks the UI in Flutter when it opens the AppAuth flow, causing Flutter UI / animations to freeze. This patch spawns AppAuth in a separate thread so that the methodCall can return immediately to Flutter, preventing the UI thread from being blocked.

### DIFF
--- a/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
+++ b/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
@@ -202,8 +202,8 @@ public class FlutterAppauthPlugin implements MethodCallHandler, PluginRegistry.A
     }
 
 
-    private void performAuthorization(AuthorizationServiceConfiguration serviceConfiguration, String clientId, String redirectUrl, ArrayList<String> scopes, String loginHint, Map<String, String> additionalParameters, boolean exchangeCode, ArrayList<String> promptValues) {
-        AuthorizationRequest.Builder authRequestBuilder =
+    private void performAuthorization(AuthorizationServiceConfiguration serviceConfiguration, String clientId, String redirectUrl, ArrayList<String> scopes, String loginHint, Map<String, String> additionalParameters, final boolean exchangeCode, ArrayList<String> promptValues) {
+        final AuthorizationRequest.Builder authRequestBuilder =
                 new AuthorizationRequest.Builder(
                         serviceConfiguration,
                         clientId,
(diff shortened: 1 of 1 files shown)


acoutts:
Note: I have only implemented and tested this fix with part of the lib. If it is a sufficient fix and there's no security concerns with running AppAuth on a separate thread, it should be applied to the other calls as well.

android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
```diff
@@ -225,14 +225,21 @@ private void performAuthorization(AuthorizationServiceConfiguration serviceConfi
             authRequestBuilder.setAdditionalParameters(additionalParameters);
         }
 
-        AppAuthConfiguration.Builder authConfigBuilder = new AppAuthConfiguration.Builder();
-        if (allowInsecureConnections) {
-            authConfigBuilder.setConnectionBuilder(InsecureConnectionBuilder.INSTANCE);
-        }
+        // Launch the AppAuth on a separate thread so that this one can immediately return
+        // to Flutter, preventing the UI from being locked up.
+        Thread t = new Thread(new Runnable() {
```
MaikuB: Thanks for the PR. From my understanding, short tasks can use `AsyncTask`, which also makes it easier to return results on the UI thread (which is how I've seen done in other plugins). `google_sign_in` plugin has their own class for executing work in the background that then returns results in the UI thread but `AsyncTask` should suffice.  If this works (note: need to ensure cancelling propagates the exception to the Flutter side) then I reckon the changes could be applied where needed. I haven't taken a look at the other methods but it's possible that only applying it to `performAuthorization`  is sufficient

acoutts: Thanks for the insight! I'll preface with I'm no expert in Android, so this is a good experience for me. One point- we technically don't ever return anything to Flutter from the method call since everything is done asynchronously via the redirect URL / deep link. Do we need to go that deep into it?

Is this simple thread approach not recommended do you think?

acoutts: Another data point - with my testing, iOS seems to handle it fine with no UI blocking in the current implementation.

MaikuB: Assuming the token call needs the change, that is one that returns a result. Also, the authorisation related methods start an activity and that's a UI related activity so should happen on the main thread. AsyncTask has a way to do this. Everything that takes place before then could happen in the background. Might be worth digging deeper to see why that causes a performance issue. I suspect it's to do with fetching with configuration info.

MaikuB:
FYI I've made a bunch of changes that involved changing the folder structure of the repository. You'll need to resolve the conflicts if you plan to work on this any further. All of the Android and iOS code is now in the flutter_appauth folder

MaikuB:
Closing this due to lack of further updates. Feel free to re-open if you proceed further

sharpsan:
This is still a stark issue.  I will try to implement what was discussed in this PR thread as soon as possible.

DASPRiD:
@sharpsan I hope you find time to revisit this PR – this is definitely a bug which needs to be fixed :)

PR #87, opened 2020-02-20
[flutter_appauth_platform_interface] Fixes error on casting the tokenAdditionParameters property from calling token method

### LINKED ISSUE #86: Cast Exception on using tokenAdditionalParameters Map
When using the method `token` of `method_channel_flutter_appauth.dart` the `tokenAdditionalParameters` attribute of the returned TokenResponse throws following Exception, as soon as I use it:

```
[VERBOSE-2:ui_dart_state.cc(157)] Unhandled Exception: type 'int' is not a subtype of type 'String' in type cast
#0      CastMap.forEach.<anonymous closure> (dart:_internal/cast.dart:288:25)
#1      _LinkedHashMapMixin.forEach (dart:collection-patch/compact_hash.dart:379:8)
#2      CastMap.forEach (dart:_internal/cast.dart:287:13)
#3      MapBase.mapToString (dart:collection/maps.dart:35:9)
#4      MapMixin.toString (dart:collection/maps.dart:192:32)
#5      _StringBase._interpolate (dart:core-patch/string_patch.dart:845:19)
```

The cause for this problem is the cast in this line https://github.com/MaikuB/flutter_appauth/blob/123010981d4d84c7addd2354c2e0e936a75c05b2/flutter_appauth_platform_interface/lib/src/method_channel_flutter_appauth.dart#L65

it should be cast to 
`result['tokenAdditionalParameters']?.cast<String, dynamic>());`.

svendroid:
Fixes https://github.com/MaikuB/flutter_appauth/issues/86

### DIFF
--- a/flutter_appauth_platform_interface/lib/src/method_channel_flutter_appauth.dart
+++ b/flutter_appauth_platform_interface/lib/src/method_channel_flutter_appauth.dart
@@ -62,6 +62,6 @@ class MethodChannelFlutterAppAuth extends FlutterAppAuthPlatform {
                 result['accessTokenExpirationTime'].toInt()),
         result['idToken'],
         result['tokenType'],
-        result['tokenAdditionalParameters']?.cast<String, String>());
+        result['tokenAdditionalParameters']?.cast<String, dynamic>());
   }
 }


MaikuB:
I've changed the title to add more details. Thanks a lot for this. I guess you're the first one to have actually used this method for since no one else had picked up on this ^^;

svendroid:
Thanks for the quick replies. I'm not really using the map, I stumbled about it because I logged all properties during development, but maybe someone else will in the future. ;)

MaikuB:
Funny enough, I had just spoken to someone about how it's great to see pull requests get submitted so it's great timing ;)

PR #109, opened 2020-04-30
Fixes null pointer exceptions when attempting to return a result for …

athornz:
…a null pending operation

This situation can occur when too many operations are started at one time. Once one operation completes, there is no longer a pending operation so any operations that complete after will throw a null pointer exception.


MaikuB:
Thanks a lot for the PR :)

PR #112, opened 2020-05-08
[flutter_appauth] & [flutter_appauth_platform] - Add support for ephemeral sessions on iOS

matthewtsmith:
As of iOS 12, with the introduction of ASWebAuthentication, Apple removed the ability for applications to clear cookies related to an SSO provider when using ASWebAuthentication. 

You can see an issue related to this with AppAuth filed here: https://github.com/openid/AppAuth-iOS/issues/402

What makes this worse on iOS is that launching a LOGOUT url in safari does not clear the cookies set from the ASWebAuthentication like it does on Android.

This causes a problem for many enterprise customers who require the ability to completely sign out of an SSO session.

This PR uses the new flag introduced in the iOS 13 to `prefersEphemeralWebBrowserSession` on the ASWebAuthenticationSession.

AppAuth supports custom user agents as noted in [this comment](https://github.com/openid/AppAuth-iOS/issues/402#issuecomment-506595068). 

For my implementation, I made a copy of `OIDExternalUserAgentIOS` which is the default userAgent for iOS and created a new one called `OIDExternalUserAgentIOSNoSSO` which sets the `prefersEphemeralWebBrowserSession` flag on iOS. The TokenRequest has a new field `prefersEphemeralSession` that defaults to `false` to maintain backwards compatibility. Setting this flag to true will cause the SDK to use the NonSSO userAgent.


flutter_appauth/ios/Classes/FlutterAppauthPlugin.m
```diff
@@ -148,14 +151,14 @@ -(void)handleAuthorizeMethodCall:(NSDictionary*)arguments result:(FlutterResult)
...
+            [self performAuthorization:configuration clientId:requestParameters.clientId clientSecret:requestParameters.clientSecret scopes:requestParameters.scopes redirectUrl:requestParameters.redirectUrl additionalParameters:requestParameters.additionalParameters preferEphemeralSession:requestParameters.preferEphemeralSession result:result exchangeCode:exchangeCode];
                                                             }];
     }
     
     
 }
 
-- (void)performAuthorization:(OIDServiceConfiguration *)serviceConfiguration clientId:(NSString*)clientId clientSecret:(NSString*)clientSecret scopes:(NSArray *)scopes redirectUrl:(NSString*)redirectUrl additionalParameters:(NSDictionary *)additionalParameters result:(FlutterResult)result exchangeCode:(BOOL)exchangeCode{
+- (void)performAuthorization:(OIDServiceConfiguration *)serviceConfiguration clientId:(NSString*)clientId clientSecret:(NSString*)clientSecret scopes:(NSArray *)scopes redirectUrl:(NSString*)redirectUrl additionalParameters:(NSDictionary *)additionalParameters preferEphemeralSession:(BOOL)preferEphemeral result:(FlutterResult)result exchangeCode:(BOOL)exchangeCode{
```
MaikuB, line 161: can you rename `preferEphemeral` to `preferEphemeralSession`?

flutter_appauth/ios/Classes/FlutterAppauthPlugin.m
```diff
@@ -167,18 +170,34 @@ - (void)performAuthorization:(OIDServiceConfiguration *)serviceConfiguration cli
...
-                                                                                   }];
+
+        NSObject<OIDExternalUserAgent> *agent = NULL;
+        if (preferEphemeral) {
+            agent = [[OIDExternalUserAgentIOSNoSSO alloc]
+                initWithPresentingViewController:rootViewController];
+        } else {
+            agent = [[OIDExternalUserAgentIOS alloc]
+                initWithPresentingViewController:rootViewController];
+        }
```
MaikuB, line 181: It's only been repeated twice but would you be able to refactor this block of code out to a method that returns the appropriate agent?

flutter_appauth/ios/Classes/OIDExternalUserAgentIOSNoSSO.h
```diff
@@ -0,0 +1,46 @@
+/*! @file OIDExternalUserAgentIOS.h
```
MaikuB, line 1: This doesn't match the name of the file. Could you rename it appropriately and include an attribution on what it's based on? This could perhaps include a link to the local of the file in the AppAuth iOS GitHub repo

flutter_appauth/ios/Classes/OIDExternalUserAgentIOSNoSSO.m
```diff
@@ -0,0 +1,253 @@
+/*! @file OIDExternalUserAgentIOS.m
```
MaikuB, line 1: See comment made in the header file

flutter_appauth/ios/Classes/OIDExternalUserAgentIOSNoSSO.m
```diff
@@ -0,0 +1,253 @@
...
+        }
+      }];
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
+      if (@available(iOS 13.0, *)) {
+          authenticationVC.presentationContextProvider = self;
+      }
+#endif
+      _webAuthenticationVC = authenticationVC;
+        if (@available(iOS 13.0, *)) {
+            authenticationVC.prefersEphemeralWebBrowserSession = @YES;
```
MaikuB, line 119: Is the `@` symbol needed here and if not could you remove it? As far as I know, it doesn't and the existing convention in the code doesn't use it either

flutter_appauth_platform_interface/lib/src/common_request_details.dart
```diff
@@ -26,6 +26,9 @@ class CommonRequestDetails implements Mappable {
   /// Whether to allow non-HTTPS endpoints (only applicable on Android)
   bool allowInsecureConnections;
 
+  /// Whether to use an ephemeral session and not share the cookies across the OS
```
MaikuB, line 29: Can you add a comment to indicate that it's specific to iOS? This can be noted in a separate line that is separate by a blank line to follow the [effective Dart documentation guidelines](https://dart.dev/guides/language/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph)

flutter_appauth/example/lib/main.dart
```diff
@@ -175,7 +175,9 @@ class _MyAppState extends State<MyApp> {
       final AuthorizationTokenResponse result =
           await _appAuth.authorizeAndExchangeCode(
         AuthorizationTokenRequest(_clientId, _redirectUrl,
-            serviceConfiguration: _serviceConfiguration, scopes: _scopes),
+            serviceConfiguration: _serviceConfiguration,
+            scopes: _scopes,
+            preferEphemeralSession: true),
```
MaikuB, line 180: Rather than forcing to be true in the existing example (which would also make it a bit out of sync with the readme), could you leave the original code as is and add another button (only visible on iOS) to the example that would then invoke the code where `preferEphemeralSession` is set to true?

matthewtsmith:
@MaikuB I've made the changes requested. Let me know if you see anything else.

A couple of notes about these changes:
1.) In the example app, I had to bump the Dart version to >= 2.5.0 in order to use control flow collections. This was the simplest way to add the iOS only button into the Column Layout

2.) I found a bug with how I was using the ArgumentProcessor. Previously I was assigning `_preferEphemeralSession` as follows

```_preferEphemeralSession = [ArgumentProcessor processArgumentValue:arguments withKey:@"preferEphemeralSession"]```

But this was assigning it to `true` simply if a value existed causing `preferEphemeralSession` to always be true.

I updated this to

```
_preferEphemeralSession = [[ArgumentProcessor processArgumentValue:arguments withKey:@"preferEphemeralSession"] isEqual:@YES];
```
and now it behaves correctly.

File: flutter_appauth/example/pubspec.yaml [hunk 8]
```diff
@@ -4,7 +4,7 @@ version: 1.0.0+1
 publish_to: 'none'
 
 environment:
-  sdk: ">=2.1.0 <3.0.0"
+  sdk: ">=2.5.0 <3.0.0"
```
MaikuB, line 7: can you make that `>=2.3.0 <3.0.0`? Support for control flow collections was actually added in 2.3

flutter_appauth/ios/Classes/FlutterAppauthPlugin.m
```diff
@@ -189,10 +193,18 @@ - (void)performAuthorization:(OIDServiceConfiguration *)serviceConfiguration cli
                 [self finishWithError:AUTHORIZE_ERROR_CODE message:[self formatMessageWithError:AUTHORIZE_ERROR_MESSAGE_FORMAT error:error] result:result];
             }
         }];
-        
     }
 }
 
+- (NSObject<OIDExternalUserAgent> *)userAgentWithEphemeralSession:(BOOL)preferEphemeralSession inRootViewController:(UIViewController *)rootViewController {
```
MaikuB, line 199: Are you able to come up with a better name for this method? Currently it implies it returns a user agent with an ephemeral session. The only idea I have at the moment is `initUserAgent`

flutter_appauth/ios/Classes/OIDExternalUserAgentIOSNoSSO.h
```diff
@@ -0,0 +1,44 @@
+/*! @file OIDExternalUserAgentIOSNoSSO.h
+    @copydetails
+        Licensed under the Apache License, Version 2.0 (the "License");
```
MaikuB, line 3: Missing attribution/link to the header file it's based of https://github.com/openid/AppAuth-iOS/blob/master/Source/iOS/OIDExternalUserAgentIOS.h

flutter_appauth/ios/Classes/OIDExternalUserAgentIOSNoSSO.m
```diff
@@ -0,0 +1,241 @@
+/*! @file OIDExternalUserAgentIOSNoSSO.m
+    @brief OIDExternalUserAgentIOSNoSSO is a custom user agent based off of the default user agent in the AppAuth iOS SDK found here:
```
MaikuB, line 2: Minor grammar issue `OIDExternalUserAgentIOSNoSSO is a custom user agent based off of...` here. `OIDExternalUserAgentIOSNoSSO is a custom user agent based on...` should work

flutter_appauth/ios/Classes/OIDExternalUserAgentIOSNoSSO.m
```diff
@@ -0,0 +1,241 @@
...
+           https://github.com/openid/AppAuth-iOS/blob/master/Source/iOS/OIDExternalUserAgentIOS.h
```
MaikuB, line 3: This file is actually based on the implementation file at https://github.com/openid/AppAuth-iOS/blob/master/Source/iOS/OIDExternalUserAgentIOS.m not the header file that is currently written here

flutter_appauth/example/lib/main.dart
```diff
@@ -75,8 +75,20 @@ class _MyAppState extends State<MyApp> {
               ),
               RaisedButton(
                 child: const Text('Sign in with auto code exchange'),
-                onPressed: _signInWithAutoCodeExchange,
+                onPressed: () => _signInWithAutoCodeExchange(),
               ),
+              if (Theme.of(context).platform == TargetPlatform.iOS)
```
MaikuB, line 80: If you look at the API docs for [`Theme.of(context).platform`](https://api.flutter.dev/flutter/material/ThemeData/platform.html), its usage isn't appropriate here. Instead you can use [`Platform.isIOS`](https://api.dart.dev/stable/2.8.2/dart-io/Platform/isIOS.html) from the `dart:io` library

flutter_appauth/example/pubspec.yaml
matthewtsmith: It was available as a feature in 2.3 but as an experimental feature, it was "officially" released in Dart 2.5. If I switch it to 2.3 I will need to add `--control-flow-collections` to the analysis_options.yaml file. 

Also, since`flutter_appauth` currently depends on flutter SDK `1.12.13+hotfix.5` or greater which bundles in Dart 2.7, I can't see a situation where using this plugin with 2.3 would be possible.

Your call though, let me know how you'd like to proceed.

flutter_appauth/ios/Classes/FlutterAppauthPlugin.m
matthewtsmith: I struggled with this name as well... I was trying to jump back into my old OjbC days where the method name is meant to read like documentation. I've decided to change it to
`- (NSObject<OIDExternalUserAgent> *)userAgentWithViewController:(UIViewController *)rootViewController useEphemeralSession:(BOOL)useEphemeralSession
`

Since creating a userAgent always requires a VC, that is moved to the first argument. I think that makes it more clear that useEphemeralSession is option you are choosing whether to enable or not.

To me, `init*` in this case seems incorrect since it's not part of a class wide setup.

matthewtsmith:
@MaikuB I've made the code updates you requested and left some comments for you to review. Please take another look when you have a chance.

For testing, the demo identity server displays the correct behavior when using the example application.

If you sign in with the `ephemeral session` button, you'll notice that.
1.) You are **not** prompted with the popup "flutter_appauth_example" Wants to use "identityserver.io"....
2.) If you tap the button a second time ( after already logging in ), you'll notice that it prompts you for your username and password again.

If you do the same test with the "Sign in with auto code exchange" button, you'll notice the second time you tap the button you're signed in without entering the u/p.

flutter_appauth/ios/Classes/OIDExternalUserAgentIOSNoSSO.h
MaikuB: @matthewtsmith missed this one :)

MaikuB: Oops sorry looks like it's there now. perhaps GitHub showed me a cached copy for some reason

MaikuB:
Nevermind about the previous question, I forgot you mentioned that trying to trigger another ephemeral session again requires you to enter the details again. That would be consistent with how it opens a private browser session that is expected to occur :)

flutter_appauth_platform_interface/lib/src/token_request.dart
```diff
@@ -15,7 +15,8 @@ class TokenRequest with CommonRequestDetails {
       String discoveryUrl,
       this.authorizationCode,
       this.codeVerifier,
-      bool allowInsecureConnections = false})
+      bool allowInsecureConnections = false,
+      bool preferEphemeralSession = false})
```
MaikuB, line 19: Apologies as I missed this last time but  believe this class should have the changes reverted as the parameter has no effect for a token request right?

matthewtsmith: @MaikuB You are correct. I've removed it from TokenRequest

MaikuB: Thanks though the platform interface tests are failing at the moment. This can be solved by moving the field from the `CommonRequestDetails` to the `AuthorizationParameters` instead. I tried to make the changes and push them but it failed. I'm guessing this is because the setting to allow for maintainers to make changes to this PR wasn't enabled

PR #115, opened 2020-05-14
[flutter_appauth] Fix crash when using logout on iOS

nytoan:
With flutter appauth, in order to do a logout we need to make an authorize call with the logout endpoint. It works well on Android but crash in iOS when creating the processed response.

This fix will check for the availability of the values before trying to set it in the processed response.


MaikuB:
Sorry for the late response and thanks for the PR. This fix may be valid for the scenario you're looking at but would end up masking issues with calling `authorize` as it's intended so will not be looking at merging this

PR #146, opened 2020-10-01
[flutter_appauth] Check if intent is null in onActivityResult

li0nza:
This seems to crash apps when intent is null.


MaikuB:
Thanks for the PR, however this is slightly different than what was discussed in the thread where an error should be reported. As it is right now, it would result in a future that never completes

li0nza:
Would this be more sufficient? Unfortunately I am not very well versed with java and this is about as much as I could figure out

PR #149, opened 2020-10-05
[flutter_appauth][flutter_appauth_platform_interface] Add untrusted SSL certificates support

### LINKED ISSUE #47: Provide support for accepting bad SSL certificates
Hi,

I'm doing some OIDC development on my local machine. Would it be possible to add an extensibility point for overriding SSL CRL / Chain validation checks? 

Best regards
Anders

josencv:
Hi! I needed to use self-signed certificates for my local and dev environments with HTTPS and it was currently not supported, so I added support for it (Android only). This PR may not compile as it is because I changed both packages, but I didin't make in the dependencies of `flutter_appauth` (read below).

Hope it helps.

-------

This PR adds support for SSL connections with untrusted certificates (meant to be used for dev environments only, with self-signed certificated). Fixes #47 and the other half of #12 (users that want a SSL connection with self-signed certificates).

**Remarks:**
- This is a breaking change as it removes the `allowInsecureConnections` flag and adds a `connectionType` enum instead with the following options: `secure` (HTTPS with trusted certificated), `untrusted` (HTTPS allowing bad certificates) and `insecure` (HTTP).
- This PR modifies both flutter_appauth and flutter_appauth_platform_interface, because the new `connectionType` argument is part of the interface contract.
- I didin't have the time to do a full test to ensure that I didin't broke anything, sorry :(
- This change is for Android only
- It looks like Android Studio 4 adds a `gradle/wrapper` directory that is meant to be added to version control, but I may be wrong.

**TODO:** (I left this undone on purpose because im not really sure which versions to use)
- [ ] Bump version in puspec.yaml in both packages
- [ ] Update `flutter_appauth` dependency on `flutter_appauth_platform_interface` so is uses the version of this PR (when I used it locally I had my local version of `flutter_appauth` depend on my local version of `flutter_appauth_platform_interface`)
- [ ] Update changelog
- [ ] Issue a warning from the plugin when using the `ConnectionType.untrusted` option, to let the developer know it shouldn't be used in production environments. To be honest, I don't really know if there is a recommended way to do logging from native code in flutter plugins.


MaikuB:
Thanks for the PR, I'll need to spend more time to look through this. My concern here is I'd need a way to test all these changes, would you happen to have any servers available that would help with this or able to direct me to some tutorials or what not that would help in getting such environments up and running easily/quickly?

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
```diff
@@ -149,6 +159,7 @@ public void onMethodCall(MethodCall call, Result result) {
         switch (call.method) {
             case AUTHORIZE_AND_EXCHANGE_CODE_METHOD:
                 try {
+                    System.out.println(arguments);
```
josencv, line 162: Sorry, this line should not have been committed. I'm going to delete it.

josencv:
> would you happen to have any servers available that would help with this or able to direct me to some tutorials or what not that would help in getting such environments up and running easily/quickly?

I'm afraid I don't have any public server to test this on  😕

MaikuB:
what about docs/tutorials i could use to get spin up servers. i'm hesitant to take more contributions on this given the issues i've seen reported and difficulty in testing this. not sure how you are testing this at the moment as well

MaikuB:
Going to close this one to clean up as it's been a number of years now with no response/info to help test this out. Happy to revisit this should anyone be able to provide some guidance

PR #160, opened 2020-11-24
[flutter_appauth] Reworked identity provider tutorial section, added FusionAuth.

mooreds:
I added another flutter tutorial which uses appauth_flutter, and reworked it so that other providers can easily add their tutorials.

Full disclosure: I'm an employee of FusionAuth. I didn't write the tutorial, but had a hand in getting it published. If this PR isn't cool to submit, no worries, just thought I'd suggest it.

### DIFF
--- a/flutter_appauth/README.md
+++ b/flutter_appauth/README.md
@@ -5,14 +5,14 @@
 
 A Flutter bridge for AppAuth (https://appauth.io) used authenticating and authorizing users. Note that AppAuth also supports the PKCE extension that is required some providers so this plugin should work with them.
 
-Provider tutorials:
-
-* [Auth0](https://auth0.com/blog/get-started-with-flutter-authentication/)
-* [FusionAuth](https://fusionauth.io/blog/2020/11/23/securing-flutter-oauth/)
-
 **IMPORTANT NOTES**:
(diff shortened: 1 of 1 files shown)


flutter_appauth/README.md
```diff
@@ -5,7 +5,10 @@
 
 A Flutter bridge for AppAuth (https://appauth.io) used authenticating and authorizing users. Note that AppAuth also supports the PKCE extension that is required some providers so this plugin should work with them.
 
-If you're using Auth0, there is a detailed post on the Auth0 blog on how you can use this plugin with Auth0 that you can check out by opening [this link](https://auth0.com/blog/get-started-with-flutter-authentication/).
+Provider tutorials:
```
MaikuB, line 8: Could this be rephrased? This is to avoid confusion with the [provider](https://pub.dev/packages/provider) package in case readers think they will be seeing how to see this plugin together with the package. Not sure what this could options there are here but may need to longer to make it clearer. Maybe "Identity provider tutorials" is enough or something like "Tutorials from identity providers". Happy to take suggestions here.

Also as it's a section in itself, using markdown to indicate this would help call that out more

MaikuB (changes requested):
Appreciate the disclosure and happy to include in the readme once some minor tweaks are done :)

flutter_appauth/README.md
mooreds: Split out the section, made it clear it wasn't related to the "provider" package. Thanks for the feedback!

MaikuB (approved):
Thanks for making the changes. I've updated the PR title for clarification since that's what will be in the commit

Issue #102, opened 2020-04-12
[Android] Token Lost if "don't keep activities" enabled


codeknox:
I was assigned this bug to investigate, and my conclusion is that it only happens when the user has enabled "don't keep activities" on Android settings.

The flow of the app is like this:

* app started
* it detects no token available
* starts oAuth
* user autorizes
* app started (again)

the flow makes sense, so I'm not here to discuss that, but I really wish someone could give me an idea of how to go around this?

I'm even thinking about the possibility of telling the user that with "don't keep activities" on, our app does not work.

Or is there a way to store the retrieved code on disk (like shared preferences) so I can check when the app starts the second time?

Thank you all for any hints,
.CodeKnox

MaikuB:
Yes your app will need to consider and be responsible for the storage of tokens as part of using OAuth 2.0,  regardless of which plugin you use for authentication. If you haven't read more about OAuth 2.0 then I would suggest reading more about it. For example in https://developers.google.com/identity/protocols/oauth2/native-app, you see it mentioned that token information needs to be kept in long lived storage so tokens can be used between multiple sessions of an app. 

Based on my understanding of what you've written as part of your application's flow, it sounds like this is missing and it doesn't check for existing tokens. This might include where an access token has expired so rather than restarting the auth process, you should be using the refresh token to get a new one.

codeknox:
Hey there @MaikuB , thank you for taking the time to write this explanation.
I understand what you wrote and we do save te token and do refresh the token.

But this is happening before I even get the token.

Let me number the steps we do:

The flow of the app is like this:

1. app started
2. it detects no token available
3. starts oAuth
4. user autorizes
5. app started (again)

As you see on step 2, we do not have a token yet.
This makes the app start the oAuth authentication process (step 3).
Once user has authorized (step 4), control is returned to our app, but because of the "Don't keep activities', app is just restarted (step 5). As we have never received the token to begin with, this becomes step 1 and the flow restarts again.

For now, the only way I see around this is to actually hack into your source code and "save" the token, so I could check on the possible return to 1.

Hope this is clearer now?
.CodeKnox

MaikuB:
Ah apologies, yes that's clearer now. Since you mentioned token storage I thought this was after the token was obtained. Does seem like you need to hack it. If you look on the repo for the Android SDK there is at least one open related issue https://github.com/openid/AppAuth-Android/issues/240

MaikuB:
Closing as there's not much I can do about this

[...]

PR #170, opened 2021-01-05
[flutter_appauth] Fix crash when authorizationService has been disposed

guykogus:
I'm not sure exactly how to fix this issue, but it seems that the `authorizationService` is disposed while the token refresh is still being called. I would've like to call `finishWithError` in this state but not sure what error to pass.

The error occurs in `flutter_appauth: ^0.9.2+6`.

```
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'void net.openid.appauth.AuthorizationService.performTokenRequest(net.openid.appauth.TokenRequest, net.openid.appauth.AuthorizationService$TokenResponseCallback)' on a null object reference
       at io.crossingthestreams.flutterappauth.FlutterAppauthPlugin.performTokenRequest(FlutterAppauthPlugin.java:353)
       at io.crossingthestreams.flutterappauth.FlutterAppauthPlugin.access$500(FlutterAppauthPlugin.java:42)
       at io.crossingthestreams.flutterappauth.FlutterAppauthPlugin$4.onFetchConfigurationCompleted(FlutterAppauthPlugin.java:278)
       at net.openid.appauth.AuthorizationServiceConfiguration$ConfigurationRetrievalAsyncTask.onPostExecute(AuthorizationServiceConfiguration.java:364)
       at net.openid.appauth.AuthorizationServiceConfiguration$ConfigurationRetrievalAsyncTask.onPostExecute(AuthorizationServiceConfiguration.java:305)
       at android.os.AsyncTask.finish(AsyncTask.java:755)
       at android.os.AsyncTask.access$900(AsyncTask.java:192)
       at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:772)
       at android.os.Handler.dispatchMessage(Handler.java:107)
       at android.os.Looper.loop(Looper.java:224)
       at android.app.ActivityThread.main(ActivityThread.java:7590)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:539)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:950)
```

### DIFF
--- a/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
+++ b/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
@@ -349,10 +349,12 @@ public class FlutterAppauthPlugin implements FlutterPlugin, MethodCallHandler, P
 
         TokenRequest tokenRequest = builder.build();
         AuthorizationService authorizationService = allowInsecureConnections ? insecureAuthorizationService : defaultAuthorizationService;
-        if (clientSecret == null) {
-            authorizationService.performTokenRequest(tokenRequest, tokenResponseCallback);
-        } else {
-            authorizationService.performTokenRequest(tokenRequest, new ClientSecretBasic(clientSecret), tokenResponseCallback);
+        if (authorizationService != null) {
+            if (clientSecret == null) {
+                authorizationService.performTokenRequest(tokenRequest, tokenResponseCallback);
+            } else {
+                authorizationService.performTokenRequest(tokenRequest, new ClientSecretBasic(clientSecret), tokenResponseCallback);
+            }
         }
 
     }


MaikuB:
Have you got a link to a repo that can reproduce this issue as I haven't seen anyone else report this. There's an example app in the repo that I typically use for testing and didn't see this issue the last time I tried it. Suspect your issue might be to do with something else as if it's do with refreshing token then I would've expected to have heard more reports on this.

guykogus:
It's for my company, so I can't share the repo. I've actually never seen the crash happen for me personally, I only saw this appear in Crashlytics and it's happening often enough that it'll be really bad when the app goes live.

MaikuB:
Understood regarding the repo though I was referring to a minimal app to reproduce the problem. As you say it'll be bad when the app goes live, I take it that it hasn't gone live yet. If so, does this mean you are running on your own devices for testing? As no one else has reported the problem, I'm wondering if it's to do with the settings on said devices. There's a setting in developer mode that doesn't keep activities. This setting will cause issues (e.g. https://github.com/MaikuB/flutter_appauth/issues/102) and there isn't much the plugin can do about it. My understanding is that is a developer only setting as well. Could this possibly have been enabled on the devices you had issues with?

guykogus:
@MaikuB the crash reports were coming from beta testing (via Firebase). The app just went live, so I was using my forked version, and there are still crashes. Need to investigate further...

gacalves:
This failure has occurred in a published application. We managed to capture through Crashlytics. 
It has occurred intermittently on devices with android 8, 9, 10 and 11.

via-guy:
@MaikuB any chance this can be merged? It really can't break anything, only fix a potential crash, no?

via-guy:
@MaikuB it looks like several people are waiting on this. Is there any reason to not merge it?

MaikuB:
Please read this comment https://github.com/MaikuB/flutter_appauth/pull/170#issuecomment-755157074

MaikuB:
For those with the issue, could you provide more details. So far I've gotten crash logs but no information on the scenario this occurs, how the plugin is being used (e.g. add-to-app or full Flutter app) etc

via-guy:
The problem is that it's really hard to reproduce and I'm using my own fork to avoid the crashes happening in production.

Even if you're cleaning up the resources as expected according to the Flutter lifecycle, it doesn't mean that no more code will run in that class. The crash is in `performTokenRequest`, which is called from several places, some of which run asynchronously (e.g. in `handleTokenMethodCall`). If the plugin is detached before that callback executes the method will still be called and this crash will occur.

Regardless of why it happens, it's obvious that it's happening, and it's not necessarily a bug with our implementation of the project.

Why is this null check is such an issue for you? In most modern languages, like Kotlin, Swift and Dart, you just add a nice little `?` and it performs the null check for you, safely executing the call. The compiler won't even build until you handle the nullability. Java doesn't let you do it in a nice way or warn you about the potential crash, but we found it and fixed it. Why refuse it?

MaikuB:
> Even if you're cleaning up the resources as expected according to the Flutter lifecycle, it doesn't mean that no more code will run in that class.

Yep I know. The point of me raising this is it would apply to all plugins not this one such that all plugins get detached that there should be more plugins having issue

> Why is this null check is such an issue for you? In most modern languages, like Kotlin, Swift and Dart, you just add a nice little ? and it performs the null check for you, safely executing the call. The compiler won't even build until you handle the nullability. Java doesn't let you do it in a nice way or warn you about the potential crash, but we found it and fixed it. Why refuse it?

This wasn't just about being hesitant about just a null check though one of the concerns I had was it could hide the root cause though what you said makes sense. The most important part I was concerned about is if you had read the comment from @guykogus who created the PR, was that they actually said they used their fork with these changes and it didn't fix the issue. With your comment a while back [here](https://github.com/MaikuB/flutter_appauth/pull/170#issuecomment-866734216), you hadn't said it had helped fix the issue either. From a general standpoint, if a code change doesn't end up affecting fixing an issue, that doesn't provide much of a reason to merge it in and no one is telling me it has helped them, this doesn't provide me much reason to merge it. If more people are actually able to come forward to help test then and had told me this had fixed the issue then I could've looked into this. Hope that made sense.

MaikuB:
@stilbu in case you haven't read my previous messages, note that it changes in this PR still don't fix the issue. To address this it's best to provide steps to reproduce and/or those in community such as yourself, verify if using a fork with the changes in this PR fixes the problem

stilbu:
> @stilbu in case you haven't read my previous messages, note that it changes in this PR still don't fix the issue. To address this it's best to provide steps to reproduce and/or those in community such as yourself, verify if using a fork with the changes in this PR fixes the problem

I know, and that is exactly what we are trying to do, but so far have not been able to reproduce this yet ourselves.

via-guy:
I really want to have this closed off so that we can go back to using the official version isn't of the one with our fix.

Let me show you a straightforward example of how this can crash:
1. A message is sent to `handleTokenMethodCall` (FlutterAppauthPlugin.java, line 291).
2. `tokenRequestParameters.serviceConfigurationParameters == null`, which means that a network call is made to retrieve configuration.
3. The plugin is disposed, for whatever reason, so `disposeAuthorizationServices()` is called
4. Retrieving configurations completes (from step 2), calling the completion callback, which calls `performTokenRequest`, which then crashes because the auth services have been disposed and set to `null`.

The root problem is that your `AuthorizationServiceConfiguration` methods are all `static`, which means that you don't have a network manager that you can cancel the calls. Or at least there's no cancel mechanism for the network calls.

So you can approve this PR and merge, or you can refactor that other class and then find a way to cancel the network calls. Either way, we can't publish our app without this fix.

karmazinkd:
we're also experiencing this crash `flutter_appauth: 4.2.0`

JordyLangen:
We are quite some of these crashes in the field for us. (currently at 11k in the past two weeks).

We are capturing all events that occur before it, and (after removing some internal events), this is always the same sequence of events:

image

Some additional details:

1. It's a full Flutter app
2. It does not happen because of a background process (we have some of those)
3. It happens on all versions of Android.

I've cloned this package into our project and made some changes (like adding a 5s delay before assigning and calling `authorizationService.performTokenRequest` ), but even then I'm failing to reproduce it.

The fact that our eventing mentions that our `ContentActivity` gets destroyed would mean the app gets killed (which does seem a bit unlikely looking at how many times it happened). If anyone has any idea on how to reproduce this, I'm more then willing to try and fix it.

via-guy:
@JordyLangen I wish I could reproduce this, but the bottom line is that my PR fixes the issue. In our app we just used my fork and that solved it. Unfortunately the author of the library would rather leave the crash than allow some defensive programming, since that doesn't pinpoint the exact cause.

MaikuB:
@via-guy note though that you hadn't actually said this fixed your issue until your last message and @guykogus who is the author of the PR has said this didn't even fix this crash. Had you and others mentioned the changes helped then I would've looked into it.  FYI your point out that what you talked about in https://github.com/MaikuB/flutter_appauth/pull/170#issuecomment-1131835418 is incorrect in that I didn't write the code for `AuthorizationServiceConfiguration` so it's not like I could look at refactoring it. That class comes from the AppAuth Android SDK. 

@JordyLangen thanks for sharing the info. At work, we've got an app using this plugin for 1-2 years now that's used for in-store staff and haven't seen these reported. I can recognise what you have came from Sentry there and what sticks out there is that app lifecycle is reported as being in the background despite how the Chrome custom tab should be redirecting back to the app. It does sound like app or at least Flutter engine is being disposed as that's the only time this plugin tries to clean up its own resources. Have you tried the following

-use Sentry to log the Flutter application lifecycle events that correspond to https://api.flutter.dev/flutter/dart-ui/AppLifecycleState.html?
- tried using the changes in this PR to see if would resolve the issue for you? Your company would likely have a bigger userbase to help with testing so if you can try it out then let me know then that would be great

JordyLangen:
> @JordyLangen I wish I could reproduce this, but the bottom line is that my PR fixes the issue. In our app we just used my fork and that solved it. Unfortunately the author of the library would rather leave the crash than allow some defensive programming, since that doesn't pinpoint the exact cause.

Doing a `null` check feels more like threating the sympton then solving the cause. It might be the only viable solution, but that is hard to determine without knowing what is causing it to happen.

> @JordyLangen thanks for sharing the info. At work, we've got an app using this plugin for 1-2 years now that's used for in-store staff and haven't seen these reported. I can recognise what you have came from Sentry there and what sticks out there is that app lifecycle is reported as being in the background despite how the Chrome custom tab should be redirecting back to the app. It does sound like app or at least Flutter engine is being disposed as that's the only time this plugin tries to clean up its own resources.

We see the crash occur when refreshing the tokens, not when signing in, so there is no custom chrome tab open (right?).

> use Sentry to log the Flutter application lifecycle events that correspond to https://api.flutter.dev/flutter/dart-ui/AppLifecycleState.html?

I'm adding more logging to get more clear.

> Tried using the changes in this PR to see if would resolve the issue for you? Your company would likely have a bigger userbase to help with testing so if you can try it out then let me know then that would be great

No, as it likely hides the problem instead of solving it.

MaikuB:
> We see the crash occur when refreshing the tokens, not when signing in, so there is no custom chrome tab open (right?).

@JordyLangen Thanks for the correction and yes my understanding is if you're only refreshing a token then it won't show a Chrome custom tab as that should only involve a HTTP request. This does however, make the issue more confusing as I thought this was due to the app perhaps killing the main activity that runs the Flutter app whilst the Chrome custom tab is open. Does your application have logic to periodically refresh the tokens based on some kind of timer? I don't have I've built but the reason I'm asking is if you happen to do so, then perhaps what is happening is that the app is trying to refresh the token and in the middle of doing it's been suspended so the resources are being released. This could explain the entries you see in Sentry and why I've not come across this myself as my personal approach has been to check if the token needs to be refreshed prior to making a request to a secured endpoint. I suspect you could try to reproduce this by following https://developer.android.com/training/monitoring-device-state/doze-standby#testing_your_app_with_app_standby

JordyLangen:
@MaikuB what I can see so far is the following:

Users who experience this issue seem to do a quick "burst" session. They open the app, perform one action, and close the app again. Usually within 1-3 seconds.

> Does your application have logic to periodically refresh the tokens based on some kind of timer?

We only refresh when we are on the foreground and when the token actually expired.

MaikuB:
In that case I believe it's worth trying to test putting the app into standby whilst it's trying to refresh. This may require forcing the refresh and using a fork so that the operation on the Android side takes longer to do in order to allow time for entering adb commands to put it into standby mode as they call it

PR #187, opened 2021-02-21
[flutter_appauth_platform_interface][flutter_appauth] migrate to null safety

MaikuB:


smiLLe:
Do you need help finishing this PR?

MaikuB:
Nope. This is left open with a prerelease as it was only done recently. I was looking to let it stay in prerelease for more time to get feedback on issues etc before it goes to stable

PR #209, opened 2021-05-11
[flutter_appauth] Fix crash when authorizationService has been disposed

via-guy:
Replaces #170


via-guy:
I just meant that I didn't need to replace #170, the changes are still the same between pre and post nnbd.

PR #216, opened 2021-06-27
[flutter_appauth][flutter_appauth_platform_interface] added support for end session requests

### LINKED ISSUE #48: Add logout
Currently there is no way to end an active session via this library.

MaikuB:
Closes #48


smiLLe:
It is working for me on Android (will test iOS tomorrow) but only when i add `idToken` and `postRedirectUrl`. If i do not add both it will redirect to the IdS, log out but somehow the session remains active

bobs4462:
Hi, I've tried out it on Android, and it kinda works: it opens the webview, logs user out, and clears all the cookies. But after that it just sits there on authentication form, and there's no way to programmatically close the webview. 
Any ideas why that might be happening?

I've tried to provide all sorts of post_logout_redirect_uri and id_token_hint, but the result is always the same, the webview is not closed automatically. And when user closes it manually, we get error `end_session_failed`.

smiLLe:
For people coming here because the logout on iOS says "Sign in".
https://github.com/openid/AppAuth-iOS/issues/643

PR #221, opened 2021-07-07
[flutter_appauth][flutter_appauth_platform_interface] add responseMode as parameter for android and ios

sbarth:

### DIFF
--- a/flutter_appauth/CHANGELOG.md
+++ b/flutter_appauth/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 1.1.0
+* Make use of responseMode inside FlutterAppauthPlugin for authorizeAndExchangeCode flow. Was needed otherwise an exception was thrown in the AuthorizationRequest.Builder when setting the responseMode via additionalParameters map.
+
 ## 1.0.0+1
 
 * There are no functional changes in this release. The only changes done were to suppress warnings that were occurring as a result of making use of Android v1 embedding APIs for backwards compatibility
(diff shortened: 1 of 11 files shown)


MaikuB:
Thanks for the PR. Looks like the build is failing as one of the tests needs to be updated to match the changes as well. Would you be able to update that?

sbarth:
Ok, fixed it finally. It was my first dart test and I didn't understand the problem ;-)
Can you take a look into flutter_appauth/pubspec.yaml, I don't know what I have to set there. Maybe I have to split this PR into 2 separate and create first PR for flutter_appauth/flutter_appauth_platform_interface and then for flutter_appauth/flutter_appauth and use the version number?

MaikuB:
Thanks. It's not necessary to split them. The reason it's not necessary is that `flutter_appauth`'s dependency on `flutter_appauth_platform_interface` is defined such that it can pull down versions newer than 3.0.0 inclusive but older than 4.0.0. This means when 3.1.0 is published, it's possible for those on that even on  version 1.0.0 of `flutter_appauth` to pull the changes that were made to the platform interface such that it's not necessary to publish a new version. However, it's better to do so to notify others of the changes done as the platform interface is a transitive dependency of the plugin.

I haven't gotten around to testing it yet but changes look straightforward so I don't anticipate there being any issues. The remaining changes I see needed for the PR are

- revert the changes done to `flutter_app_appauth`'s pubspec.yaml on how the dependency on the platform interface is defined
- to bump the version of `flutter_appauth_platform_interface` to 3.1.0 and add a changelog entry.
- to bump `flutter_app_appauth` to 1.1.0 with a changelog entry on it too

Are the last two something you're ok with pushing to this PR? Otherwise I could also do this provided you had enabled the setting that allows edits from maintainers.

sbarth:
- [x] revert the changes done to flutter_app_appauth's pubspec.yaml on how the dependency on the platform interface is defined 
- [x] to bump the version of flutter_appauth_platform_interface to 3.1.0 and add a changelog entry.
- [x] to bump flutter_app_appauth to 1.1.0 with a changelog entry on it too 

All ok for me :-)

MaikuB:
Thanks. I've amended the changelog a bit and added `responseMode` as a parameter to `AuthorizationRequest` as well

PR #228, opened 2021-07-30
[flutter_appauth] Skip https issuer check if allowInsecureConnections is true.

rfuerst87:
Appauth fails to verify the id token and throws an `AuthorizationException` if you try to authorize against an auth service over http even if `allowInsecureConnections` is true. This PR simply sets the `skipIssuerHttpsCheck` flag to true if `allowInsecureConnections` is true.


MaikuB:
Thanks for this!

PR #253, opened 2021-10-30
[flutter_appauth] added macosx platform interface implementation

noliran:
Hi folks,

Thank you for `flutter_appauth`! While I'm not that familiar with Objective C, I found it relatively easy to take the iOS platform interface and use it to create a MacOS platform interface. The changes are small and a side-by-side comparison is feasible.

Note that `AppAuth-iOS` supports two different methods to receive the authorization response redirect:
1. Custom URI scheme
2. Built-in HTTP server

Due to my lack of familiarity with Objective C, I've refrained from adding support for the second method (built-in HTTP server), and have only implemented custom URI scheme handling. I've taken the custom URI handling code from `https://github.com/openid/AppAuth-iOS/blob/master/Examples/Example-macOS/Source/AppDelegate.m`

Usage instructions:
* Add the following to `macos/Runner/Info.plist`
```
    <key>CFBundleURLTypes</key>
    <array>
        <dict>
            <key>CFBundleTypeRole</key>
            <string>Editor</string>
            <key>CFBundleURLSchemes</key>
            <array>
                <string>customschemename</string>
            </array>
        </dict>
    </array>
```
* Don't forget to upgrade your `flutter_secure_storage` (if you use it) to the 5.x branch (which adds support for MacOS), and to add the required Keychain Sharing capability through Xcode


MaikuB:
Thanks for the PR. Is the code for implementing the macOS side the same as what's in iOS though? I would imagine so as the AppAuth iOS and macOS plugin share code and if that is indeed what you did, be best to see if both the iOS and macOS side could share the relevant code files so fixes/enhancements to iOS or macOS don't require duplicating the code.

Note: i'll be trying to see if I can release the end session changes first

North101:
This PR still works with the latest code so it'd be great if it could be merged

MaikuB:
@North101 what do you mean by latest code? Copying the latest objective C code and use that for macOS? If so the comment I raised has yet to be addressed and file sharing is meant to be possible. Copying and pasting code for every update isn't ideal and adds more maintenance burden not just for myself but others that submit PRs. If macOS support is of interest to you then you could contribute by looking at a PR that addresses this and I'd be happy to close this in favour of new one

JordyLangen:
I've looked into a solution using symlinks to avoid duplication. The `performAuthorization` method uses classes from `UIKit` which makes this not as straightforward as you might hope.

MaikuB:
Closing this in favour of #320

MaikuB:
Closing this in favour of #320

PR #259, opened 2021-11-11
Support ui_locales parameters

dimitristoz:
Added check in additionalParameters if there is ui_locales parameters, and set it in android with the new function which has been added at the last version of appAuth plugin otherwise application crashes

### DIFF
--- a/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
+++ b/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
@@ -307,6 +307,11 @@ public class FlutterAppauthPlugin implements FlutterPlugin, MethodCallHandler, P
             authRequestBuilder.setResponseMode(responseMode);
         }
 
+        if(additionalParameters != null && !additionalParameters.isEmpty() && additionalParameters.containsKey("ui_locales")){
+            authRequestBuilder.setUiLocales(additionalParameters.get("ui_locales"));
+            additionalParameters.remove("ui_locales");
+        }
+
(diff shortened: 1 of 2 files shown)


flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
```diff
@@ -307,6 +307,11 @@ private void performAuthorization(AuthorizationServiceConfiguration serviceConfi
             authRequestBuilder.setResponseMode(responseMode);
         }
 
+        if(additionalParameters != null && !additionalParameters.isEmpty() && additionalParameters.containsKey("ui_locales")){
```
MaikuB, line 310: The first two conditions of this `if` statement are duplicated in the next `if` statement that existed prior to this PR. Would you be able to refactor the code to remove the duplication?

MaikuB (changes requested):
Thanks for the PR. I left a comment around refactoring. Would you be be able to do make that change?

MaikuB:
Also I updated the PR title as this isn't specific to keycloak

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
dimitristoz: Sure i ll change it

dimitristoz: I pushed one more commit with changes.

MaikuB (approved):

Issue #255, opened 2021-11-05
PlatformException(token_failed, Failed to get token: [error: null, description: Invalid ID Token], null, null)


postpasspost:
Anyone has any idea of what this error means ?

It works on iOS, but not on Android. Same configuration files 

I updated from old version of Flutter.

```
flutter_appauth: ^1.1.0+2

environment:
  sdk: ">=2.7.0 <3.0.0"
```

I did a Flutter cache clean

And here is the code
```
  final TokenResponse tokenResponse = await flutterAppAuth.token(
    TokenRequest(
        clientId,
        redirectUri,
        clientSecret: clientSecret,
        serviceConfiguration: serviceConfiguration,
        authorizationCode: authorizationResponse.authorizationCode,
        codeVerifier: authorizationResponse.codeVerifier),
  );
```

```
I/flutter ( 9705): login error: PlatformException(token_failed, Failed to get token: [error: null, description: Invalid ID Token], null, null) -> #0      StandardMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:607:7)
I/flutter ( 9705): #1      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:156:18)
I/flutter ( 9705): <asynchronous suspension>
I/flutter ( 9705): #2      MethodChannelFlutterAppAuth.token (package:flutter_appauth_platform_interface/src/method_channel_flutter_appauth.dart:53:9)
I/flutter ( 9705): <asynchronous suspension>
I/flutter ( 9705): #3      login (package:NAME/utils/auth.dart:72:39)
I/flutter ( 9705): <asynchronous suspension>
I/flutter ( 9705): #4      _LoginState.loginAction (package:NAME/login.dart:199:5)
I/flutter ( 9705): <asynchronous suspension>
I/flutter ( 9705): #5      _LoginState.build.<anonymous closure>.<anonymous closure>.<anonymous closure> (package:NAME/login.dart:164:29)
I/flutter ( 9705): <asynchronous suspension>

```

MaikuB:
It would mean that the underlying Android AppAuth SDK has found an invalid ID token per the error message. This is the link to the repo for the SDK where you can go through its code and error messages https://github.com/openid/AppAuth-Android as this is something you'd need to dig further specific to your scenario

MaikuB:
[...]

PR #265, opened 2021-12-01
[flutter_appauth] Return detailed error message to flutter

akaegi:
See issue https://github.com/MaikuB/flutter_appauth/issues/255

### DIFF
--- a/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
+++ b/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
@@ -310,7 +310,6 @@ public class FlutterAppauthPlugin implements FlutterPlugin, MethodCallHandler, P
                         clientId,
                         ResponseTypeValues.CODE,
                         Uri.parse(redirectUrl));
-//        authRequestBuilder.setNonce(null);
 
         if (scopes != null && !scopes.isEmpty()) {
             authRequestBuilder.setScopes(scopes);
--- a/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
+++ b/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
@@ -159,7 +159,7 @@ public class FlutterAppauthPlugin implements FlutterPlugin, MethodCallHandler, P
                     checkAndSetPendingOperation(call.method, result);
                     handleAuthorizeMethodCall(arguments, true);
                 } catch (Exception ex) {
-                    finishWithError(AUTHORIZE_AND_EXCHANGE_CODE_ERROR_CODE, ex.getLocalizedMessage());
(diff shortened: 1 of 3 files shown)


flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
```diff
@@ -310,6 +310,8 @@ private void performAuthorization(AuthorizationServiceConfiguration serviceConfi
                         clientId,
                         ResponseTypeValues.CODE,
                         Uri.parse(redirectUrl));
+//        authRequestBuilder.setNonce(null);
```
MaikuB, line 313: As this is commented out, could you remove this instead?

akaegi: Oups, sorry for this left-over 😐

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
```diff
@@ -431,7 +432,9 @@ private void performEndSessionRequest(AuthorizationServiceConfiguration serviceC
     }
 
     private void finishWithTokenError(AuthorizationException ex) {
-        finishWithError(TOKEN_ERROR_CODE, String.format(TOKEN_ERROR_MESSAGE_FORMAT, ex.error, ex.errorDescription));
+        final Throwable cause = ex.getCause();
+        final String causeDescription = cause != null ? cause.getMessage() : null;
+        finishWithError(TOKEN_ERROR_CODE, String.format(TOKEN_ERROR_MESSAGE_FORMAT, ex.errorDescription, causeDescription));
```
MaikuB, line 437: Dropping the use of the `error` property of the `AuthorizationException` could lead to less information about what error happened in other scenarios. if you look at the API docs for this property (https://github.com/openid/AppAuth-Android/blob/ceb253b8118e481f3fe6f648772a8d16179a11fe/library/java/net/openid/appauth/AuthorizationException.java#L620), it's meant for errors that are part of OAuth 2 protocol. There's also code within AppAuth Android SDK that populates it https://github.com/openid/AppAuth-Android/blob/ceb253b8118e481f3fe6f648772a8d16179a11fe/library/java/net/openid/appauth/AuthorizationService.java#L655

With this in mind, I would say it's better to cater for all scenarios by first reverting the code to what it was before so that the `error` is still returned as it did before and pass the `causeDescription` to the third parameter of the method call at https://github.com/MaikuB/flutter_appauth/blob/49d7919e53fba8a7372884485b12f2849089b502/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java#L447 so that it's populates the [details](https://api.flutter.dev/flutter/services/PlatformException/details.html) property of the `PlatformException` on the Flutter side

MaikuB:
As I hadn't seen updates on this for a while, I made the changes mentioned myself based on what you originally had

akaegi:
Thanks for taking over! Sorry, the issue slipped away from my todo list 😐

MaikuB:
A PR would still be welcome on making sure the nonce can be returned and specified :)

Issue #291, opened 2022-01-23
Passing claims in AuthorizationRequestion additionalParameters causes crash on Android


garry-jeromson:
The `claims` parameter is an optional part of the OpenID standard when making an authorization request (see https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter).

Currently, this parameter is not supported by the Flutter interface layer, so the only way to pass it is via the `additionalParameters` argument when constructing an `AuthorizationRequest`. This, however, leads to the following error on Android:

```
FATAL EXCEPTION: main
E/AndroidRuntime( 7453): Process: ch.some.app, PID: 7453
E/AndroidRuntime( 7453): java.lang.IllegalArgumentException: Parameter claims is directly supported via the authorization request builder, use the builder method instead
E/AndroidRuntime( 7453): 	at net.openid.appauth.Preconditions.checkArgument(Preconditions.java:132)
E/AndroidRuntime( 7453): 	at net.openid.appauth.AdditionalParamsProcessor.checkAdditionalParams(AdditionalParamsProcessor.java:62)
E/AndroidRuntime( 7453): 	at net.openid.appauth.AuthorizationRequest$Builder.setAdditionalParameters(AuthorizationRequest.java:1050)
E/AndroidRuntime( 7453): 	at io.crossingthestreams.flutterappauth.FlutterAppauthPlugin.performAuthorization(FlutterAppauthPlugin.java:336)
E/AndroidRuntime( 7453): 	at io.crossingthestreams.flutterappauth.FlutterAppauthPlugin.access$400(FlutterAppauthPlugin.java:42)
E/AndroidRuntime( 7453): 	at io.crossingthestreams.flutterappauth.FlutterAppauthPlugin$2.onFetchConfigurationCompleted(FlutterAppauthPlugin.java:262)
E/AndroidRuntime( 7453): 	at net.openid.appauth.AuthorizationServiceConfiguration$ConfigurationRetrievalAsyncTask.onPostExecute(AuthorizationServiceConfiguration.java:417)
E/AndroidRuntime( 7453): 	at net.openid.appauth.AuthorizationServiceConfiguration$ConfigurationRetrievalAsyncTask.onPostExecute(AuthorizationServiceConfiguration.java:358)
E/AndroidRuntime( 7453): 	at android.os.AsyncTask.finish(AsyncTask.java:771)
E/AndroidRuntime( 7453): 	at android.os.AsyncTask.access$900(AsyncTask.java:199)
[...]

PR #292, opened 2022-01-23
[flutter_appauth] Handle OpenID claims parameter correctly for authorization requests on Android

garry-jeromson:
A fix for https://github.com/MaikuB/flutter_appauth/issues/291#issue-1111841681

### DIFF
--- a/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
+++ b/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
@@ -22,6 +22,9 @@ import net.openid.appauth.TokenRequest;
 import net.openid.appauth.TokenResponse;
 import net.openid.appauth.connectivity.DefaultConnectionBuilder;
 
+import org.json.JSONException;
+import org.json.JSONObject;
+
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Map;
@@ -51,6 +54,7 @@ public class FlutterAppauthPlugin implements FlutterPlugin, MethodCallHandler, P
     private static final String TOKEN_ERROR_CODE = "token_failed";
     private static final String END_SESSION_ERROR_CODE = "end_session_failed";
     private static final String NULL_INTENT_ERROR_CODE = "null_intent";
+    private static final String INVALID_CLAIMS_ERROR_CODE = "invalid_claims";
 
     private static final String DISCOVERY_ERROR_MESSAGE_FORMAT = "Error retrieving discovery document: [error: %s, description: %s]";
     private static final String TOKEN_ERROR_MESSAGE_FORMAT = "Failed to get token: [error: %s, description: %s]";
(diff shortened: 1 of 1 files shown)


flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
```diff
@@ -333,6 +336,12 @@ private void performAuthorization(AuthorizationServiceConfiguration serviceConfi
                 authRequestBuilder.setUiLocales(additionalParameters.get("ui_locales"));
                 additionalParameters.remove("ui_locales");
             }
+
+            if(additionalParameters.containsKey("claims")){
+                final JSONObject claimsAsJson = new JSONObject(additionalParameters.get("claims"));
```
MaikuB, line 341: Looks like the pipeline indicates that Android app is failing to build though due to a potential unhandled `JSONException`. Would it be possible to catch this and return an error over to the Dart side? How this is done elsewhere within the Java code is by calling the `finishWithError` method.

MaikuB (changes requested):
Thanks for the PR. Left a comment there on one are of the code.

Correct me if I'm wrong but other minor thing from looking the PR title and content is that it now looks like the PR is now only relevant to `flutter_appauth`. From what I've read, using `additionalParameters` on iOS should work without any additional changes

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
garry-jeromson: Thanks for the hint - updated accordingly. :-)

garry-jeromson: Can also confirm that using `additionalParameters` on iOS works like a charm.

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
```diff
@@ -333,6 +337,17 @@ private void performAuthorization(AuthorizationServiceConfiguration serviceConfi
...
             }
+
+            if(additionalParameters.containsKey("claims")){
+                try {
+                    final JSONObject claimsAsJson = new JSONObject(additionalParameters.get("claims"));
+                    authRequestBuilder.setClaims(claimsAsJson);
+                    additionalParameters.remove("claims");
+                }
+                catch (JSONException ex) {
+                    finishWithError(INVALID_CLAIMS_ERROR_CODE, ex.getLocalizedMessage(), getCauseFromException(ex));
```
MaikuB, line 348: Given how the code is written, if an exception occurs, it'll be caught and dealt with but will continue to remaining logic in this method. You'd need to add a `return` statement to prevent this from happening. The `finishWithError()` call is more to notify the Flutter/Dart about the progress of the method

MaikuB (commented):
Thanks for making the change. One more minor tweak needed

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
garry-jeromson: Good catch - will make it so.

MaikuB:
Thanks for this. Android app isn't compiling since the Flutter SDK deprecated Android v1 embedding and isn't an issue with your PR so will merge this in so that fixing example app can be addressed separately

MaikuB (approved):

PR #301, opened 2022-02-16
[flutter_appauth] Skip HTTPS check for unsecure connections for attributes built 'onAttachedToEngine' method.

jpuerto:
This PR is to finish the fix started with the PR #228. This fix works fine for authorization, but on token refresh the HTTPS check error is triggered again.


MaikuB:
Good spot and thanks for the PR!

PR #307, opened 2022-02-25
[flutter_appauth] fix prompt delimit with space

angle319:
Hi MaikuB,
I found a bug for using multiple parameter "prompt". Should delimit with space.

image

Refer [OIDC spec](https://openid.net/specs/openid-connect-core-1_0-27.html#AuthRequest).


MaikuB:
Thanks!

PR #310, opened 2022-03-02
[flutter_appauth] Fix minor typos in README

ryanjohndias:
I've been reading over the docs quite a bit recently, and keep noticing these two typos. Thought I'd save newcomers a few seconds of confusion when copy-pasting the code block results in a compiler fail 🙂


MaikuB:
Thanks :)

PR #318, opened 2022-04-06
[flutter_appauth][flutter_appauth_platform_interface] Use default system browser outside app

onizuka87:
Hi,

For one of my clients, I needed to use the default system browser outside of the app, therefore I've added a `(bool) defaultSystemBrowser` property to `AuthorizationParameters` and `EndSessionRequest` which uses `[OIDExternalUserAgentIOSCustomBrowser CustomBrowserSafari]` in the `(id<OIDExternalUserAgent>)userAgentWithViewController` method.

This change is only for iOS.

### DIFF
--- a/flutter_appauth/example/lib/main.dart
+++ b/flutter_appauth/example/lib/main.dart
@@ -91,6 +91,18 @@ class _MyAppState extends State<MyApp> {
                         preferEphemeralSession: true),
                   ),
                 ),
+              if (Platform.isIOS)
+                Padding(
+                  padding: const EdgeInsets.all(8.0),
+                  child: ElevatedButton(
+                    child: const Text(
+                      'Sign in with auto code exchange using default system browser (iOS only)',
+                      textAlign: TextAlign.center,
+                    ),
+                    onPressed: () =>
+                        _signInWithAutoCodeExchange(defaultSystemBrowser: true),
+                  ),
+                ),
               ElevatedButton(
                 child: const Text('Refresh token'),
                 onPressed: _refreshToken != null ? _refresh : null,
@@ -216,7 +228,8 @@ class _MyAppState extends State<MyApp> {
   }
 
   Future<void> _signInWithAutoCodeExchange(
-      {bool preferEphemeralSession = false}) async {
+      {bool preferEphemeralSession = false,
+      bool defaultSystemBrowser = false}) async {
     try {
       _setBusyState();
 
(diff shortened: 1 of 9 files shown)


MaikuB:
Thanks for the PR. I got a couple of questions

- Whilst this behaves as expected, is there some documentation on how AppAuth iOS SDK is able to determine the default browser? Asking as I couldn't find any and it looks confusing to me given it makes use of `CustomBrowserSafari` yet the browser used could actually, say, Chrome if it was set to to be the default browser
- What happens if the user closes the browser i.e. cancels the login request? The demo IdentityServer instance seems to be down but it seems as though with this approach that the future never completes. Is this what you find happening as well? If so, given you seem to have an app making use of this, how would you suggest others deal with this scenario?

The other thing that concerns me is regarding the naming of the parameter and its default value. Suppose I don't have any other browser besides Safari, in this case to me it makes sense that the value of `systemDefaultBrowser` is `true` and that it defaults to `true` as Safari is used behind the scenes.. In this case, IMO it's really more about whether or not the browser is opened outside of the app or not. This to me implies that the parameter could have a better name to describe what it does. I'm not sure what would make for a good name here though but the ones that come to mind would something like `opensDefaultBrowserExternally` or `opensInExternalDefaultBrowser`

joelle-supy-io (approved):

MaikuB:
Closing this as there's no further response and the second point I raised would cause an issue with a future that never completes

Kuurse:
I'd really like to see #613 merged so I'll do my best to answer your question about the future not resolving: 

 - If a user doesn't complete the login process for whatever reason, the Future indeed never completes. This can be handled by keeping track of the login attempts and tracking the app lifecycle with the `didChangeAppLifecycleState` method to see if the user returned to the app. This could be used in conjunction with a timeout: 
 `final result = await FlutterAppAuth()
          .authorizeAndExchangeCode(request)
          .timeout(const Duration(minutes: 2));`
          
I could post a small example if you would like.

MaikuB:
@Kuurse yep if you can share an example then that would much appreciated. I'm also open to a new PR based off #613 that includes the work done there with an updated example app with what you mentioned. If possible, these changes should only apply to a scenario where a custom browser is used

Kuurse:
I'm unsure what you're asking here. Do you want me to fork the code from #613 and add a usage example? If yes, I haven't brought any changes to his code so would just an example suffice?

MaikuB:
Yep, I know that it would have been unlikely that you would have made changes to the code in #613 so I am referring to adding a usage example by adding to [this example app](https://github.com/MaikuB/flutter_appauth/tree/master/flutter_appauth/example)

PR #320, opened 2022-04-22
[flutter_appauth] Add support for macOS

JordyLangen:
Add macOS support to `flutter_appauth`.

Implementation summary:

- Moved specific iOS and macOS logic into specific impl. classes
- Load specific platform impl using targetting
- All Classes in the `macos` directory are symlinks to the needed classes in `ios`

I haven't used Objective-C in ages, and I'm an Android & Flutter developer. Feel free to adjust any code you see.


JordyLangen:
Inspired by #253

MaikuB:
Thanks so much for this. Do you think you'd be able to update the PR so that the example app can be run on macOS and that the cirrus script (https://github.com/MaikuB/flutter_appauth/blob/master/.cirrus.yml) also has a task to try to build the example app on macOS? Here's a reference of how I've done this on another plugin https://github.com/MaikuB/flutter_local_notifications/blob/master/.cirrus.yml#L30

JordyLangen:
@MaikuB Sure no problem. I'm not familiar with cirrus, any hints or pointers?

MaikuB:
Take a look at the links I gave in the previous post. The only thing missing is an update to melos.yaml file and here's a reference https://github.com/MaikuB/flutter_local_notifications/blob/b187373310718926cf649999686a87a479fceeda/melos.yaml#L47

I believe you should be about to copy and paste the relevant snippets and it should just work

JordyLangen:
Hey @MaikuB , I'm currently a bit swamped at work. I'm intending to make sure your comments are addressed, but it will not be fast sadly. Feel free to make any changes you want yourself if you'd like to see this merged earlier.

MaikuB:
@JordyLangen i've updated the PR to add the parts I mentioned. Other things changed are

- addressed missing implementation for end session request
- addressed missing implementation for ephemeral sessions. this required defining a "no SSO" agent like the one that exists for iOS
- removed symbolic links for the macOS classes and moved them to the macos folder as it's looks like there was no need for it

would you be able to take a look and provide some feedback, especially on the last point? not sure if i've missed a reason for having the macOS classes in the iOS folder

JordyLangen:
@MaikuB looks all good 👍  I simply followed an approach I saw in google repositories where they tackled the "shared logic" in a similar fashion.

MaikuB:
Thanks a lot for taking a look and for the PR itself :)

PR #321, opened 2022-04-23
[flutter_appauth]feat: add state parameter

Sembauke:

### DIFF
--- a/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
+++ b/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
@@ -204,6 +204,7 @@ public class FlutterAppauthPlugin implements FlutterPlugin, MethodCallHandler, P
         final String discoveryUrl = (String) arguments.get("discoveryUrl");
         final String redirectUrl = (String) arguments.get("redirectUrl");
         final String loginHint = (String) arguments.get("loginHint");
+        final String state = (String) arguments.get("state");
         clientSecret = (String) arguments.get("clientSecret");
         final ArrayList<String> scopes = (ArrayList<String>) arguments.get("scopes");
         final ArrayList<String> promptValues = (ArrayList<String>) arguments.get("promptValues");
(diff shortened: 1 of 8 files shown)


Sembauke:
Hi @MaikuB,

I have been working on adding the state parameter.
More info on it here: https://auth0.com/docs/secure/attack-protection/state-parameters

I am not sure how to test it.

if you have any suggestions or changes you would like me to make. Let me know.

MaikuB:
Thanks for the PR. Do you mean to say you made the changes without verifying that they work? One thing I should call out is the native SDKs will generate a random state value already so this PR is really more on if apps want to do so on there own. With this in mind, I suggest that the API docs for the property gets updated to indicate a null value means the native SDKs will deal with state

Sembauke:
Thanks for the review Maiku. I am having trouble testing it in the example app as this newly added state parameter is not showing up. 

Our app needs a custom state set, so it will retrieve more personal data.

MaikuB:
This is a monorepo that can be used with [melos](https://melos.invertase.dev/) so once you've done the setup covered in their docs, melos will do some setup behind the scenes so that local dependencies are used. Without this, the code for the plugin just references the dependencies that have been published

MaikuB:
Note I've tried testing your changes and it doesn't work on iOS with a state mismatch error being returned. The state coming back from the server looks like one generated by the AppAuth iOS SDK so it looks like you are missing some changes to make sure the state value specified is actually be used

Sembauke:
Alright thanks for the review. I will look into it.

Sembauke:
Hey @MaikuB,

I am trying out Melos - It works great.  Only I am unable to do reach the demo.identityserver.io for some reason.
Also the build seems to pass on Android. But not on IOS. 

I know it has something todo with the mismatch error. Only I can't find a good way to debug. The only thing that is coming up is:

```batch
melos run build:example_ios
   └> melos exec -c 1 -- \  "flutter build ios --no-codesign"
       └> FAILED
ScriptException: The script build:example_ios failed to execute
```

MaikuB:
You'll need to update your branch with the latest changes from master. I recently pushed changes as the details of the demo IdentityServer instance changed.

As for your iOS build issue, that would be something you'd likely need to look further into as it's likely an issue with your environment. CI is setup on the repo using the same commands and you should that find the iOS build task works fine on your PR. 

For debugging, you don't really need to run the build commands to begin with. As mentioned earlier, once you bootstrap melos it'll set all projects in your clone to use local dependencies. After this you debug an app as you normally would i.e. don't need to use the melos CLI anymore after bootstrapping. If you happen to still struggle with this then you could manually update all the pubspec.yaml files to point to your local versions of each package. See https://docs.flutter.dev/development/packages-and-plugins/using-packages#dependencies-on-unpublished-packages

Sembauke:
I will not be continuing this. We found a better way of doing thing using our own API.

PR #329, opened 2022-05-13
[flutter_appauth]:bug: iOS & Android - Nonce Parameter Fix

beYsSoftware:
Hi @MaikuB, this PR is for fix the bug when we use the Nonce Parameter in the class AuthorizationTokenRequest

### DIFF
--- a/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
+++ b/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
@@ -39,6 +39,7 @@ import io.flutter.plugin.common.MethodChannel;
 import io.flutter.plugin.common.MethodChannel.MethodCallHandler;
 import io.flutter.plugin.common.MethodChannel.Result;
 import io.flutter.plugin.common.PluginRegistry;
+import android.text.TextUtils;
 
 /**
  * FlutterAppauthPlugin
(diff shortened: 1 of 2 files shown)


MaikuB:
Thanks for PR but may need to park this one or close it as I recently started working on an issue some others reported but had tried to contribute a fix too. It revolves around an ID token validation error where there's nonce mismatch. This would involve needing the plugin return the nonce value as a property and then allow it to be specified as in the token request. For consistency, both scenarios should have nonce as a property rather than being in the `additionalParameters` as done in this PR. Hope this makes sense will credit you when change goes out your PR highlight the need to use a different method on iOS to pass the nonce

MaikuB:
Thanks for PR but may need to park this one or close it as I recently started working on an issue some others reported but had tried to contribute a fix too. It revolves around an ID token validation error where there's nonce mismatch. This would involve needing the plugin return the nonce value as a property and then allow it to be specified as in the token request. For consistency, both scenarios should have nonce as a property rather than being in the `additionalParameters` as done in this PR. Hope this makes sense will credit you when change goes out your PR highlight the need to use a different method on iOS to pass the nonce

MaikuB:
Oh wait I noticed your PR tackles a different thing around authorisation requests so different to the issue I mentioned. After I've merged in #331, do you think you could rebase and update the PR based on those changes? I know your PR was up first but the issue #331 addresses has been going for a while and based on posts I've seen here, would seem to have more impact and priority

PR #349, opened 2022-07-12
[flutter_appauth][flutter_appauth_platform_interface] Add preferEphemeralSession option to logout

ziegler-daniel:
Hi,

at the moment it is possible to pass the option `preferEphemeralSession: true` to the sign in call, but not to the end session call. Therefore, before every logout a pop-up like this is shown to the user:
image

If no pop-up was displayed before the login, it would be nicer if there is also none before the logout. So, I added the functionality to set the `preferEphemeralSession` option also for the end session call.

It would be nice if this feature could be included into the library.

### DIFF
--- a/flutter_appauth/README.md
+++ b/flutter_appauth/README.md
@@ -113,8 +113,8 @@ to sign in and sign out.
 
 With an ephemeral session there will be no warning like `"app_name" Wants to Use "domain_name" to Sign In` on iOS.
 
-The option `preferEphemeralSession = true` must only be used for the end session call, if it is also used for the sign in call. 
-Otherwise there will be still an active login session in the browser.
+The option `preferEphemeralSession = true` must only be used for the end session call if it is also used for the sign in call. 
+Otherwise, there will be still an active login session in the browser.
 
 ## Android setup
 
--- a/flutter_appauth/README.md
+++ b/flutter_appauth/README.md
@@ -106,6 +106,16 @@ await appAuth.endSession(EndSessionRequest(
 
 The above code passes an `AuthorizationServiceConfiguration` with all the endpoints defined but alternatives are to specify an `issuer` or `discoveryUrl` like you would with the other APIs in the plugin (e.g. `authorizeAndExchangeCode()`).
 
+### Ephemeral Sessions (iOS and macOS only)
+On iOS (versions 13 and above) and macOS you can use the option `preferEphemeralSession = true` to start an 
+[ephemeral browser session](https://developer.apple.com/documentation/foundation/urlsessionconfiguration/1410529-ephemeral)
+to sign in and sign out.
+
+With an ephemeral session there will be no warning like `"app_name" Wants to Use "domain_name" to Sign In` on iOS.
+
+The option `preferEphemeralSession = true` must only be used for the end session call, if it is also used for the sign in call. 
+Otherwise there will be still an active login session in the browser.
+
 ## Android setup
 
 Go to the `build.gradle` file for your Android app to specify the custom scheme so that there should be a section in it that look similar to the following but replace `<your_custom_scheme>` with the desired value
--- a/flutter_appauth_platform_interface/lib/src/end_session_request.dart
+++ b/flutter_appauth_platform_interface/lib/src/end_session_request.dart
@@ -40,7 +40,7 @@ class EndSessionRequest with AcceptedAuthorizationServiceConfigurationDetails {
   /// Whether to use an ephemeral session that prevents cookies and other browser data being shared with the user's normal browser session.
   ///
   /// This property is only applicable to iOS versions 13 and above.
-  bool? preferEphemeralSession;
+  bool preferEphemeralSession;
 
   final Map<String, String>? additionalParameters;
 }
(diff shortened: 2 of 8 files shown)


flutter_appauth_platform_interface/lib/src/end_session_request.dart
```diff
@@ -36,5 +37,10 @@ class EndSessionRequest with AcceptedAuthorizationServiceConfigurationDetails {
   /// This property is only applicable to Android.
   bool allowInsecureConnections;
 
+  /// Whether to use an ephemeral session that prevents cookies and other browser data being shared with the user's normal browser session.
+  ///
+  /// This property is only applicable to iOS versions 13 and above.
+  bool? preferEphemeralSession;
```
MaikuB, line 43: Can this made non-nullable?

ziegler-daniel: Yes, I changed it accordingly

MaikuB:
Whilst the popup is removed, using an ephemeral session means a "private" browser is used. Users won't actually get signed out as a result. The following is a recording from sigining in and signing out normally using the example

https://user-images.githubusercontent.com/25263378/179742937-13f7d3ab-f223-4599-b3e3-c79077dc3c7f.mov

At the end I attempt to sign in again to show that the browser no longer knows I've been signed in as I'm prompted to sign in again. Now let's repeat this with the end session request where `preferEphemeralSession` is set to `true`


https://user-images.githubusercontent.com/25263378/179743538-224fdab3-ec44-4de0-97b8-decdd99a38fc.mov

After signing out, I've tried to sign in again and still recognises that I'm signed in. This effectively means the end session request hasn't really done anything. With this is in mind and do let me know if I've missed something, I don't see a reason why this feature added by the PR since the user's session is being kept.

dconlisk:
Ah, that's a shame that it doesn't work. @MaikuB do you know if it's possible to update the text in the popup from "Sign In" to "Sign Out" when ending the session?

dconlisk:
@MaikuB I was afraid that might be the case - thanks!

ziegler-daniel:
@MaikuB thank you for your response. In my opinion the option `preferEphemeralSession = true` must only be used for the logout call, if it was also used for the sign in call. According to the [apple documentation](https://developer.apple.com/documentation/foundation/urlsessionconfiguration/1410529-ephemeral) no data is stored in an ephemeral session. So, if you use `preferEphemeralSession = true` for sign in, you won't get the described problem. Using different values for `preferEphemeralSession` for sign in and end session call would be like using two different browsers for the two calls.

The end session call invalidates the current refresh token regardless of the `preferEphemeralSession` option.

I can add this to the documentation to prevent misunderstandings.

MaikuB:
@ziegler-daniel ok gotcha. Updating the documentation around using this makes sense to me

ziegler-daniel:
> 

Sure, I added some documentation at the `EndSessionRequest` and in the readme file. Is that okay with you?

flutter_appauth/README.md
```diff
@@ -106,6 +106,16 @@ await appAuth.endSession(EndSessionRequest(
...
 The above code passes an `AuthorizationServiceConfiguration` with all the endpoints defined but alternatives are to specify an `issuer` or `discoveryUrl` like you would with the other APIs in the plugin (e.g. `authorizeAndExchangeCode()`).
 
+### Ephemeral Sessions (iOS and macOS only)
+On iOS (versions 13 and above) and macOS you can use the option `preferEphemeralSession = true` to start an 
+[ephemeral browser session](https://developer.apple.com/documentation/foundation/urlsessionconfiguration/1410529-ephemeral)
+to sign in and sign out.
+
+With an ephemeral session there will be no warning like `"app_name" Wants to Use "domain_name" to Sign In` on iOS.
+
+The option `preferEphemeralSession = true` must only be used for the end session call, if it is also used for the sign in call.
```
MaikuB, line 116: the comma isn't needed before the "if"

flutter_appauth/README.md
```diff
@@ -106,6 +106,16 @@ await appAuth.endSession(EndSessionRequest(
...
 
+### Ephemeral Sessions (iOS and macOS only)
+On iOS (versions 13 and above) and macOS you can use the option `preferEphemeralSession = true` to start an 
+[ephemeral browser session](https://developer.apple.com/documentation/foundation/urlsessionconfiguration/1410529-ephemeral)
+to sign in and sign out.
+
+With an ephemeral session there will be no warning like `"app_name" Wants to Use "domain_name" to Sign In` on iOS.
+
+The option `preferEphemeralSession = true` must only be used for the end session call, if it is also used for the sign in call. 
+Otherwise there will be still an active login session in the browser.
```
MaikuB, line 117: in this case, a comma should be after "Otherwise" :)

MaikuB (changes requested):
Yes placing the documentation like you have is fine though I left comments on the usage of commas :)

ziegler-daniel:
> Yes placing the documentation like you have is fine though I left comments on the usage of commas :)

Thank you for you feedback. I adjusted it.

MaikuB (approved):

fbernaly:
@ziegler-daniel: Thanks for this PR. This is what I needed for my app.

Issue #394, opened 2022-12-14
[Proposal] Expose better error code


timnew:
## Problem

Currently, when `AppAuth` failed, the plugin throws `PlatformException`.
The `code` of `PlatformException` is hardcoded to a certain value based on which method the app called, which isn't very useful. As the app should have know which API is called. Instead the actual useful information isn't passed up.

For example, if app calls `authorizeAndExchangeCode`, and then user canceled the web view popped up.

On iOS:
```
PlatformException(authorize_and_exchange_code_failed, Failed to authorize: The operation couldn’t be completed. (org.openid.appauth.general error -3.), null, null)
```

The code `authorize_and_exchange_code_failed` is useless as we have known we called `authorizeAndExchangeCode`.

The useful bit is actually `org.openid.appauth.general` and `-3`, which is the `code` and `domain` of the `NSError` underneath, but it is mixed in the message that passed up.

on Android:
```
Error: PlatformException(authorize_and_exchange_code_failed, Failed to authorize: [error: null, description: User cancelled flow], null, null)
```

Again `authorize_and_exchange_code_failed` is not helpful. `User cancelled flow` is the one useful. And we missed some extra information here.

## Proposal

When throw PlatformException from plugin, rather than returns hard coded thing that bind to the calling method. Expose the more detail from underneath error.

For iOS:
Use the concat the `NSError`'s `domain` and `code` as the PlatformException error code. So it would return `org.openid.appauth.general -3` instead of `authorize_and_exchange_code_failed` as code.

And for Android:
The AppAuth would throws `AuthorizationException`, which contains `type` and `code`. return the `type` and `code` accordingly.

MaikuB:
A similar thing has been mentioned before and I was mentioning that PRs are welcome on it. I'd likely not get it myself so if you can contribute then that would be great. One concern here is if both platforms even use the same error code. If they don't then this becomes messy as it requires either the consumer of the plugin to handle platform-specific error codes or have a layer within the plugin to map error platform's error code to something consistent across all platforms. The latter could take a fair bit of work to do and requires keeping an eye on new errors added to be releases of the AppAuth iOS and Android SDKs

MaikuB:
[...]

PR #395, opened 2022-12-14
[flutter_appauth] Expose native exception type and code to the Flutter code

timnew:
An implementation for the issue https://github.com/MaikuB/flutter_appauth/issues/394

It employs a new code format: `<existing code>:<domain/type>:<code>`

This might need to be considered as Breaking Changes, as it breaks the code that checks code right now. But it can be easily fixed:

For existing code:

```dart
try { 
  // calling appauth
} on PlatformException catch(ex) {
  if (ex.code == "authorize_and_exchange_code_failed") {
    // handle error
  }
}
```
It can be upgraded as 
```dart
try { 
  // calling appauth
} on PlatformException catch(ex) {
  if (ex.code.startsWith("authorize_and_exchange_code_failed:")) {
    // handle error
  }
}
```

For user cancelled the webview for auth, the error could be 
* iOS: `authorize_and_exchange_code_failed:org.openid.appauth.general:-3`
* Android: `authorize_and_exchange_code_failed:GENERAL_ERROR:1`

### DIFF
--- a/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
+++ b/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
@@ -629,7 +629,7 @@ public class FlutterAppauthPlugin
                             finishWithSuccess(tokenResponseToMap(resp, authResponse));
                         } else {
                             finishWithError(ex, AUTHORIZE_AND_EXCHANGE_CODE_ERROR_CODE,
-                                    AUTHORIZE_AND_EXCHANGE_CODE_ERROR_MESSAGE_FORMAT);
+                                    AUTHORIZE_ERROR_MESSAGE_FORMAT);
                         }
                     }
                 };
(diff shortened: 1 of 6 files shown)


MaikuB:
Hey @timnew, would it be possible to change this PR so the there's now a new Exception class that gets thrown by the plugin? The thinking behind this is to avoid having consumers of the plugin fiddling with the string and have the native error code by the native SDKs be behind a separate property. This would also be in-line with what I read recently in https://github.com/flutter/flutter/wiki/Contributing-to-Plugins-and-Packages#platform-exception-handling that IMO makes sense to adopt in this situation. This would look something like defining a new class in the platform interface package for the exception whose structure is similar to that defined in https://api.flutter.dev/flutter/services/PlatformException-class.html but have a nullable error property named something like `nativeAppAuthSDKErrorCode` whose purpose is to hold the error codes thrown by the native AppAuth Android/iOS SDKs. Let me know what you think

timnew:
@MaikuB I'll try to do it asap! Got a bit snowed by the daily work

neokree:
@timnew any updates on this MR? Could you look into it?

MewanWA:
When is this going to be completed/merged? Its been close to 4 months since there were some changes requested.

prabuddaFernando:
https://docs.github.com/articles/what-are-the-different-access-permissions

MaikuB:
Reminder that a good number of people working on open source software do so via their spare time. If you have need for something then you can look at carrying on the work yourself via your own fork

MaikuB:
Closing this in favour of #509 that has been merged in

PR #412, opened 2023-03-29
[flutter_appauth] code challenge explanation added in the example class

davideravasi:
The class used for the documentation example https://pub.dev/packages/flutter_appauth is not mentioning if the code_challenge needs to be created by who is consuming the library or if this happens under the hood. The comment explain that is the library which takes care of it and gives a brief explanation of why is not explicitly returned/used later in the flow


flutter_appauth/example/lib/main.dart
```diff
@@ -210,6 +210,9 @@ class _MyAppState extends State<MyApp> {
     try {
       _setBusyState();
       // use the discovery endpoint to find the configuration
+      // the code_challenge is generated from the code verifier and only the code verifier is included in the result
```
MaikuB, line 213: A few things
- as the comment that was there in the previous line before is no longer the only comment for code below, could you switch to a using [multi-line comments](https://dart.dev/language/comments#multi-line-comments)?
- are you able to add a full stop to break up the various sentences?
- are you able to update the comment you've added so it's consistent with how it expresses terms from RFC? There's a mix with how words are expressed (i.e. code_challenge, code verifier vs codeVerifier and authorizationCode). My suggestion would be to express this more like plain English, that is use a space to separate. In this case it would be to use "code challenge", "code verifier" and "authorisation code" (or "authorization code" if you're going by American English)

MaikuB:
Thanks for the PR. Left comments around the changes, hope they make sense. The builds are failing due to another reason that I'll fix up in a separate PR

MaikuB:
This [PR](https://github.com/MaikuB/flutter_appauth/pull/413) has been merged now to fix the build issue so if you update your fork/branch then this should resolve the errors around the failed checks

flutter_appauth/example/lib/main.dart
```diff
@@ -209,7 +209,12 @@ class _MyAppState extends State<MyApp> {
   Future<void> _signInWithNoCodeExchange() async {
     try {
       _setBusyState();
-      // use the discovery endpoint to find the configuration
+      /* 
+         The discovery endpoint (_discoveryUrl) is used to find the configuration.
+         The code challenge generation can be checked here > https://github.com/MaikuB/flutter_appauth/search?q=challenge.
+         The code challenge is generated from the code verifier and only the code verifier is included in the result.
+         This because to get the token in the method _exchangeCode (see above) we need only the code verifier and the authorization code according to the spec https://www.rfc-editor.org/rfc/rfc7636 page 9 section 4.5.
```
MaikuB, line 216: Thanks for making the change. I opened this an IDE and saw this column was fairly long that I needed to horizontally scroll to see the rest of the comments. Are you able to make each line of comment follow the line length of at most 80 characters (i.e. make it consistent the default for Flutter code)?

davideravasi:
So,
I tried to make my comment more readable.
Just a remark, at line 281 there is a comment very long that can be shrink in multiple lines if we want to follow this kind of "readable" approach.
Let me know if now is ok

MaikuB:
Thanks for letting me know. I'll fix the other comment in a separate PR

davideravasi:
Thank you for merging it @MaikuB 
Just to understand, this page > https://pub.dev/packages/flutter_appauth/example
Will be automatically updated with my comment in the next hours / day I imagine, or is there something else I need to do to make the comment appearing in the example on pub.dev? 🤔

MaikuB:
Pub.dev updates when new releases are published

PR #416, opened 2023-04-05
[various updated API docs in platform interface to avoid lines longer than 80 characters and fix docs in example

MaikuB:
Follow-on from https://github.com/MaikuB/flutter_appauth/pull/415 due to forgetting to push changes to platform interface


davideravasi:
Sorry to bother you, I see in the example class https://github.com/MaikuB/flutter_appauth/blob/master/flutter_appauth/example/lib/main.dart

That in my comment line 217 is missing an "e" (should be the) and line 221 should be spec (missing c).

Just in case you missed it ;)

MaikuB:
Sorted now, thanks for picking that up

Issue #168, opened 2021-01-02
Android appAuth: ActivityNotFound exception crash when browser is disabled/missing from user device.


urbanjagodic:
AppAuth issue: https://github.com/openid/AppAuth-Android/issues/459

When user tries to log in using **flutter_appauth plugin** and doesn't have a browser on his device or its disabled,
the application crashes with the following error:

```java
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: PID: 5758
    android.content.ActivityNotFoundException
        at net.openid.appauth.AuthorizationService.prepareAuthorizationRequestIntent(AuthorizationService.java:372)
        at net.openid.appauth.AuthorizationService.getAuthorizationRequestIntent(AuthorizationService.java:267)
        at net.openid.appauth.AuthorizationService.getAuthorizationRequestIntent(AuthorizationService.java:296)
        at io.crossingthestreams.flutterappauth.FlutterAppauthPlugin.performAuthorization(FlutterAppauthPlugin.java:314)
        at io.crossingthestreams.flutterappauth.FlutterAppauthPlugin.access$300(FlutterAppauthPlugin.java:42)
        at io.crossingthestreams.flutterappauth.FlutterAppauthPlugin$2.onFetchConfigurationCompleted(FlutterAppauthPlugin.java:234)
        at net.openid.appauth.AuthorizationServiceConfiguration$ConfigurationRetrievalAsyncTask.onPostExecute(AuthorizationServiceConfiguration.java:364)
        at net.openid.appauth.AuthorizationServiceConfiguration$ConfigurationRetrievalAsyncTask.onPostExecute(AuthorizationServiceConfiguration.java:305)
        at android.os.AsyncTask.finish(AsyncTask.java:755)
        at android.os.AsyncTask.access$900(AsyncTask.java:192)
        at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:772)
        at android.os.Handler.dispatchMessage(Handler.java:107)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)
```

Would be possible to catch the exception and instead, open up a dialog pop-up which informs the 
[...]

PR #465, opened 2023-12-19
[flutter_appauth] Catch native ActivityNotFoundException

NikHomann:
Currently, if there are no browser installed on the device, the underlying plugin AppAuth-Android will throw an ActivityNotFoundException which crashes the Flutter application. This was for instance reported here: https://github.com/MaikuB/flutter_appauth/issues/168

You can reproduce the error by running the example app on any emulator without any installed browsers (disable the default Chrome in Settings - Apps - Chrome) and trying to login.

This fixes the error by catching the ActivityNotFoundException and returning a PlattformException.

### DIFF
--- a/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
+++ b/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
@@ -348,8 +348,13 @@ public class FlutterAppauthPlugin implements FlutterPlugin, MethodCallHandler, P
         }
 
         AuthorizationService authorizationService = allowInsecureConnections ? insecureAuthorizationService : defaultAuthorizationService;
-        Intent authIntent = authorizationService.getAuthorizationRequestIntent(authRequestBuilder.build());
-        mainActivity.startActivityForResult(authIntent, exchangeCode ? RC_AUTH_EXCHANGE_CODE : RC_AUTH);
+
+        try {
+            Intent authIntent = authorizationService.getAuthorizationRequestIntent(authRequestBuilder.build());
+            mainActivity.startActivityForResult(authIntent, exchangeCode ? RC_AUTH_EXCHANGE_CODE : RC_AUTH);
+        } catch (Exception ex) {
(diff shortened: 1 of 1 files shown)


MaikuB:
Thanks for the PR. Will see how it goes once I reproduce this but initial thought from looking at the change is that the change involves catch any exception as oppose to `ActivityNotFoundException` that you mentioned. If the only cause of an `ActivityNotFoundException` is to do with a lack of browser then I would say it would better to catch that specific exception and translate this to provide more details that the API call failed due to a lack of a browser

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
```diff
@@ -348,8 +348,13 @@ private void performAuthorization(AuthorizationServiceConfiguration serviceConfi
         }
 
         AuthorizationService authorizationService = allowInsecureConnections ? insecureAuthorizationService : defaultAuthorizationService;
-        Intent authIntent = authorizationService.getAuthorizationRequestIntent(authRequestBuilder.build());
-        mainActivity.startActivityForResult(authIntent, exchangeCode ? RC_AUTH_EXCHANGE_CODE : RC_AUTH);
+
+        try {
+            Intent authIntent = authorizationService.getAuthorizationRequestIntent(authRequestBuilder.build());
+            mainActivity.startActivityForResult(authIntent, exchangeCode ? RC_AUTH_EXCHANGE_CODE : RC_AUTH);
+        } catch (Exception ex) {
```
MaikuB, line 355: Was able to reproduce this based on the steps you mentioned and did more research online where the API docs for AppAuth Android SDK confirms that it's thrown only when no suitable browser is found.  With that in mind, would you be able to make a change so that only the `ActivityNotFoundException` and introduce a new error code (e.g. `NO_BROWSER_AVAILABLE`)  and message (e.g. `No suitable browser is available`) that would that make it more easier to catch this specific issue on the Flutter/Dart side?

MaikuB (changes requested):

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
NikHomann: That makes sense, changed that.

MaikuB (approved):

Issue #459, opened 2023-10-13
Crash during redirect  after logout is successfull


rajath-1990:
I am using this library for authorization app Azure AD. Login is working fine. App is crashing after logout. Crash is happening when it redirects back to the app.  

_postLogoutRedirectUrl will be a normal website URL that has been verified by uploading asset_link.json in the server (This is required for app linking)

Please help me in fixing this.

Here is the code:

    await _appAuth.endSession(EndSessionRequest(
          idTokenHint: _idToken,
          postLogoutRedirectUrl: _postLogoutRedirectUrl,
          serviceConfiguration: _serviceConfiguration));

Crash:  at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:807)
E/AndroidRuntime( 4223): Caused by: java.lang.NullPointerException: dataIntent must not be null
E/AndroidRuntime( 4223): 	at net.openid.appauth.Preconditions.checkNotNull(Preconditions.java:55)
E/AndroidRuntime( 4223): 	at net.openid.appauth.EndSessionResponse.fromIntent(EndSessionResponse.java

android manifest: 

 <application android:label="flutter_appauth_example" android:icon="@mipmap/ic_launcher">
        <activity android:name=".MainActivity" android:launchMode="singleTop" android:theme="@style/LaunchTheme" android:configChanges="orientation|keyboardHidden|keyboard|screenSize|locale|layoutDirection|fontScale|screenLayout|density" android:hardwareAccelerated="true" android:windowSoftInputMode="adjustResize">

            <intent-filter>
                <action android:name="android.intent.action.MAIN"/>
                <category android:name="android.intent.category.LAUNCHER"/>
            </intent-filter>

            <intent-filter android:autoVerify="true">
                <action android:name="android.intent.action.VIEW"/>
                <category android:name="android.intent.category.DEFAULT"/>
                <category android:name="android.intent.category.BROWSABLE"/>
                <data android:scheme="https" android:host="somewebsite.com"/>
            </intent-filter>
        </activity>
        
        <meta-data android:name="flutterEmbedding" android:value="2"/>
    </application>

MaikuB:
[...]

PR #468, opened 2024-01-19
[flutter_appauth] Fix for logout crash while app is on background

JohnKim7:
on Android Ii user does logout and while logging out user closes the browser and go background and come back, the app crashes. It happens because flutter activity waits for intent not to be null, but it is null. We need to add null check on activity result method on Android. 

Related issue: https://github.com/MaikuB/flutter_appauth/issues/459#issuecomment-1865485929

### DIFF
--- a/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
+++ b/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
@@ -493,15 +493,20 @@ public class FlutterAppauthPlugin implements FlutterPlugin, MethodCallHandler, P
             return true;
         }
         if (requestCode == RC_END_SESSION) {
-            final EndSessionResponse endSessionResponse = EndSessionResponse.fromIntent(intent);
-            AuthorizationException ex = AuthorizationException.fromIntent(intent);
-            if (ex != null) {
-                finishWithEndSessionError(ex);
+            if (intent == null) {
+                finishWithError(NULL_INTENT_ERROR_CODE, NULL_INTENT_ERROR_FORMAT, null);
             } else {
-                Map<String, Object> responseMap = new HashMap<>();
-                responseMap.put("state", endSessionResponse.state);
-                finishWithSuccess(responseMap);
+                final EndSessionResponse endSessionResponse = EndSessionResponse.fromIntent(intent);
+                AuthorizationException ex = AuthorizationException.fromIntent(intent);
+                if (ex != null) {
+                    finishWithEndSessionError(ex);
+                } else {
+                    Map<String, Object> responseMap = new HashMap<>();
+                    responseMap.put("state", endSessionResponse.state);
+                    finishWithSuccess(responseMap);
+                }
             }
+            return true;
         }
         return false;
     }


JohnKim7:
on react native version app auth, there is such kind of null check. And as it says it prevents from app crash.

https://github.com/FormidableLabs/react-native-app-auth/pull/352

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
```diff
@@ -479,7 +479,7 @@ private String getCauseFromException(Exception ex) {
 
     @Override
     public boolean onActivityResult(int requestCode, int resultCode, Intent intent) {
-        if (pendingOperation == null) {
+        if (pendingOperation == null || intent == null) {
```
MaikuB, line 482: Thanks for the PR but I believe the check should be done further below. What's done here doesn't align with how onActivityResult works for Flutter plugins that override `onActivityResult` and what's done is results in a different approach to what's been done in the React Native AppAuth library. Flutter provides `onActivityResult` that returns a boolean so that plugins can indicate if they were the plugin that was meant to process the result (reference: https://api.flutter.dev/javadoc/io/flutter/plugin/common/PluginRegistry.ActivityResultListener.html). As you've put this check here, this results in the method returning `false` i.e. it would mean that plugin ends up reporting that it didn't handle the result. This would be incorrect and could end mean the intent is passed onto other plugins. Furthermore, it could cause the`Future` associated with the end session Dart API to not complete i.e. the app can hang on calling the end session API. This is because plugins should be using the [Result](https://api.flutter.dev/javadoc/io/flutter/plugin/common/MethodChannel.Result.html) object to report a success or error. Returning early here prevents this from happening. 

Where `intent` check should be done further is when the plugin is processing the result to do with an end session request. There's an existing "template" on this being done for authorisation checks at https://github.com/MaikuB/flutter_appauth/blob/f999c01d6e1811baf85b9a6b55ca2dfa89ce71f1/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java#L486 so I believe correct solution is to mimic that and put the check at https://github.com/MaikuB/flutter_appauth/blob/f999c01d6e1811baf85b9a6b55ca2dfa89ce71f1/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java#L496 where the plugin has confirmed that the result is associated with the request code for the end session request. Doing so would be similar to have the React Native plugin calls `promise.reject(...)` and brings consistency to how this plugin was already addressing this for authorisation requests

MaikuB: Hmm I just realised that the code block for dealing with the end session request isn't returning `true`. Would you be able to make that change as well as part of the PR? If so, then that would be much appreciated

JohnKim7: yeah, just noticed as well. I made a fix for that. You can take a look

MaikuB: LGTM, will merge it in once the checks are done. Thanks again :)

PR #474, opened 2024-01-23
[flutter_appauth] Update README.md

davideravasi:
Added information about how to handle the possible data leak on iOS due to the storage of the token in the cache.db file
Follow up of a previous PR > https://github.com/MaikuB/flutter_appauth/pull/463

### DIFF
--- a/flutter_appauth/README.md
+++ b/flutter_appauth/README.md
@@ -185,7 +185,7 @@ Go to the `Info.plist` for your iOS/macOS app to specify the custom scheme so th
 </array>
 ```
 
-Note: iOS Apps generate a file called `cache.db` which contains the table `cfurl_cache_receiver_data`. This table will contain the access token obtained after the login is completed. If the potential data leak represents a thread for your application you can disable the information caching for the entire iOS App (ex. https://kunalgupta1508.medium.com/data-leakage-with-cache-db-2d311582cf23).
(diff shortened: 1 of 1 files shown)


flutter_appauth/README.md
```diff
@@ -185,6 +185,8 @@ Go to the `Info.plist` for your iOS/macOS app to specify the custom scheme so th
 </array>
 ```
 
+Note: iOS Apps generate a file called `cache.db` which contains the table `cfurl_cache_receiver_data`. This table will contain the access token obtained after the login is completed. If the potential data leak represents a thread for your application you can disable the information caching for the entire iOS App (ex. https://kunalgupta1508.medium.com/data-leakage-with-cache-db-2d311582cf23).
```
MaikuB, line 188:
```suggestion
Note: iOS apps generate a file called `cache.db` which contains the table `cfurl_cache_receiver_data`. This table will contain the access token obtained after the login is completed. If the potential data leak represents a threat for your application then you can disable the information caching for the entire iOS app (ex. https://kunalgupta1508.medium.com/data-leakage-with-cache-db-2d311582cf23).
```

Made the suggestions based on the following
- capitalising of the "a" in "apps" is odd as if it wasn't shortened then it would be written as "iOS application(s)" so this be consistent with that
- I assume you meant to write "threat" instead of "thread"
- Added a "then" as it allows sentence to "connect" what the reader can do more naturally as the sentence is long. Not too dissimilar to writing code in a `IF/GIVEN  ... THEN ...` format

davideravasi:
Thanks for merging it! I close my other PR

PR #509, opened 2024-07-16
[flutter_appauth, flutter_appauth_platform_interface] Expose Platform errors

SamCosta1:
This PR will hopefully replace https://github.com/MaikuB/flutter_appauth/pull/395.

This change will expose errors raised by the underlying AppAuth SDKs such that they can be consumed by Dart code.
The changes are designed such that they should be entirely backwards compatible and any existing error handling should be unaffected. 

As documented in the readme errors can now be processed as follows

```
try {
  await appAuth.authorize(...);
} on FlutterAppAuthUserCancelledException catch (e) {
  // Handle user cancelling the flow (only applies to authorize and authorizeAndExchangeCode)
} on FlutterAppAuthPlatformException catch (e) {
  final FlutterAppAuthPlatformErrorDetails details = e.details;
  // Handle exceptions based on errors from AppAuth.
} catch (e) {
  // Handle unexpected errors.
}
```

Where the `FlutterAppAuthPlatformErrorDetails` class contains all the specific information exposed by the platform. 
The `FlutterAppAuthUserCancelledException` has been added since it's a common use case to need to handle this differently to other errors.


retendo:
I would love to see something like this merged. More granular handling of those errors is required in many applications. 👍 

So if I would want to handle the raw OAuth 2.0 standard errors now like "invalid_grant", "expired_token", etc. that are now provided by both the iOS and Android plugin in `FlutterAppAuthPlatformErrorDetails.error`, I can just do something like this, right?

```
try {
  await appAuth.token(...);
} on FlutterAppAuthPlatformException catch (e) {
  final String oauthError = e.details.error;
  if (oauthError != null && (oauthError == "expired_token" || oauthError == "invalid_grant" || oauthError == "invalid_token")) {
    logout();
  }
}
```

Neat! Two thumbs up!

SamCosta1:
Yes exactly, the logout use case was the driver for creating this PR 😎😎.

@MaikuB would it be possible to merge this? I'm happy to make changes if needed

MaikuB:
I'll review and take a look when I have the time to do so. This is something I maintain in my spare time and isn't the only plugin I maintain either. I suggest others point to your fork in the meantime so they can provide feedback and raise and issues should they find any. 

From the quick look I did take though, I agree that this would replace #395. Whilst I would personally like to see errors map to something that is cross platform, I understand that could be a difficult/time consuming task. In the absence of that, it looks like this PR provides a better approach than what was in #395. What I'm curious about as well though is suppose this plugin migrated to use pigeon, how would it affect this PR? Not saying I've got definite plans to do so but was something that crossed my mind. Would you happen to know?

MaikuB:
Other thing I forgot to add is I noticed is the exception for cancelling is documented to only be for `authorizeAndExchange()`. If this is true then there's a gap there as getting the authorisation code is really the only part that requires a user to sign in. Exchange the code for a token is a web request to a server's `token` endpoint.

SamCosta1:
I did think about creating a cross platform enum but as you say it quickly looked like a big task that would also take a while to test. I'm not familiar with Objective C and my Java in rusty (is there appetite for swapping to Swift?). 

I'm afraid I have absolutely no idea about Pigeon though, I've not used it before.

You're absolutely correct about the docs, I'll update those :) Thanks for your response, appreciate you're busy!

retendo:
> Whilst I would personally like to see errors map to something that is cross platform, I understand that could be a difficult/time consuming task. In the absence of that, ...

I mean, the OAuth 2 error codes are "cross platform" in a way. They are not "polluted" by client code and they are already quite useful.

We could have a sealed class for them and provide them along the error details instead of the raw error data.
Something like this:

```dart
sealed class OAuth2Error {
  final String error;
  final String? errorDescription;
  final String? errorUri;

  const OAuth2Error({
    required this.error,
    this.errorDescription,
    this.errorUri,
  });

  static OAuth2Error fromErrorCode(String error, {String? errorDescription, String? errorUri}) {
    switch (error) {
      case 'invalid_request':
        return InvalidRequest(errorDescription: errorDescription, errorUri: errorUri);
      case 'unauthorized_client':
        return UnauthorizedClient(errorDescription: errorDescription, errorUri: errorUri);
      case 'access_denied':
        return AccessDenied(errorDescription: errorDescription, errorUri: errorUri);
      case 'unsupported_response_type':
        return UnsupportedResponseType(errorDescription: errorDescription, errorUri: errorUri);
      case 'invalid_scope':
        return InvalidScope(errorDescription: errorDescription, errorUri: errorUri);
      case 'server_error':
        return ServerError(errorDescription: errorDescription, errorUri: errorUri);
      case 'temporarily_unavailable':
        return TemporarilyUnavailable(errorDescription: errorDescription, errorUri: errorUri);
      case 'invalid_client':
        return InvalidClient(errorDescription: errorDescription, errorUri: errorUri);
      case 'invalid_grant':
        return InvalidGrant(errorDescription: errorDescription, errorUri: errorUri);
      case 'unsupported_grant_type':
        return UnsupportedGrantType(errorDescription: errorDescription, errorUri: errorUri);
      case 'invalid_token':
        return InvalidToken(errorDescription: errorDescription, errorUri: errorUri);
      case 'insufficient_scope':
        return InsufficientScope(errorDescription: errorDescription, errorUri: errorUri);
      default:
        return UnknownError(error: error, errorDescription: errorDescription, errorUri: errorUri);
    }
  }
}

class InvalidRequest extends OAuth2Error {
  const InvalidRequest({super.errorDescription, super.errorUri})
      : super(error: 'invalid_request');
}

class UnauthorizedClient extends OAuth2Error {
  const UnauthorizedClient({super.errorDescription, super.errorUri})
      : super(error: 'unauthorized_client');
}

class AccessDenied extends OAuth2Error {
  const AccessDenied({super.errorDescription, super.errorUri})
      : super(error: 'access_denied');
}

class UnsupportedResponseType extends OAuth2Error {
  const UnsupportedResponseType({super.errorDescription, super.errorUri})
      : super(error: 'unsupported_response_type');
}

class InvalidScope extends OAuth2Error {
  const InvalidScope({super.errorDescription, super.errorUri})
      : super(error: 'invalid_scope');
}

class ServerError extends OAuth2Error {
  const ServerError({super.errorDescription, super.errorUri})
      : super(error: 'server_error');
}

class TemporarilyUnavailable extends OAuth2Error {
  const TemporarilyUnavailable({super.errorDescription, super.errorUri})
      : super(error: 'temporarily_unavailable');
}

class InvalidClient extends OAuth2Error {
  const InvalidClient({super.errorDescription, super.errorUri})
      : super(error: 'invalid_client');
}

class InvalidGrant extends OAuth2Error {
  const InvalidGrant({super.errorDescription, super.errorUri})
      : super(error: 'invalid_grant');
}

class UnsupportedGrantType extends OAuth2Error {
  const UnsupportedGrantType({super.errorDescription, super.errorUri})
      : super(error: 'unsupported_grant_type');
}

class InvalidToken extends OAuth2Error {
  const InvalidToken({super.errorDescription, super.errorUri})
      : super(error: 'invalid_token');
}

class InsufficientScope extends OAuth2Error {
  const InsufficientScope({super.errorDescription, super.errorUri})
      : super(error: 'insufficient_scope');
}

class UnknownError extends OAuth2Error {
  const UnknownError({required super.error, super.errorDescription, super.errorUri});
}
```

```dart
// Usage example (in plugin)
details.oauthError = OAuth2Error.fromErrorCode(
  errorCode,
  errorDescription: errorDescription,
  errorUri: errorUri
);
```

```dart
// Usage example (in app code)
try {
  await appAuth.token(...);
} on FlutterAppAuthPlatformException catch (e) {
  switch (e.details.oauthError) {
    case InvalidRequest():
    case UnauthorizedClient():
    case AccessDenied():
    case UnsupportedResponseType():
    case InvalidScope():
    case ServerError():
    case TemporarilyUnavailable():
    case InvalidClient():
      print('Something went wrong');
    case InvalidGrant():
    case UnsupportedGrantType():
    case InvalidToken():
    case InsufficientScope():
    case UnknownError():
      logout();
  }
}
```

SamCosta1:
That looks pretty good to me! I'll probably also include the raw error too though alongside this since there are other errors the different platforms spit out which aren't mappable. 

I'll add this in as soon as I get time

MaikuB:
> I mean, the OAuth 2 error codes are "cross platform" in a way. They are not "polluted" by client code and they are already quite useful.

To clarify, I was referring to the error codes within the AppAuth iOS and Android SDKs. Each SDK has different error codes for same scenarios. As the plugin is a cross-platform library, ideally developers should not have to concern themself with platform-specific error codes where possible. [This](https://github.com/openid/AppAuth-Android/blob/master/library/java/net/openid/appauth/AuthorizationException.java) is for Android whilst [this](https://github.com/openid/AppAuth-iOS/blob/master/Sources/AppAuthCore/OIDError.h) is for iOS. One example that can spotted there is is an invalid discovery doc uses `0` on Android and `-2` on iOS

MaikuB:
@SamCosta1 the response was specifically to @retendo as I suspect my point/preference from an ideal standpoint wasn't understood. It's not blocker so that's why I mentioned to you I understand it's a time consuming task. The main issue with #395 was the lack of follow-up to address the feedback I did raise and from what I've seen so far, your PR addresses it.

Agreed that what @retendo could be used to address the ones tied to RFC though I've not been to see if there's a better way that has been recommended and followed by the Flutter team. I once had a link in one of comments in #395 but it appears the page doesn't exist anymore or got relocated

retendo:
@MaikuB Thanks for clarifying, I suspected that you meant the platform specific library error, but I wanted to point out that IMO they are not that important to make a huge improvement for most cases.

I think apart from the RFC cases, the most important one is the cancellation failure case which @SamCosta1 already implemented. At least in my iOS app, which uses AppAuth-iOS, this is the only one apart from the RFC errors that I handle. That seems to be enough for a lot of use cases, so this PR is already a big improvement that also leaves room for future iteration.

flutter_appauth/example/lib/main.dart
```diff
@@ -182,11 +188,11 @@ class _MyAppState extends State<MyApp> {
       _setBusyState();
       final TokenResponse? result = await _appAuth.token(TokenRequest(
           _clientId, _redirectUrl,
-          refreshToken: _refreshToken, issuer: _issuer, scopes: _scopes));
+          refreshToken: _refreshToken! + "a", issuer: _issuer, scopes: _scopes));
```
MaikuB, line 191: I suspect this was an unintentional change (?)

flutter_appauth/example/pubspec.yaml
```diff
@@ -14,6 +14,10 @@ dependencies:
     path: ../
   http: ^0.13.0
 
+dependency_overrides:
```
MaikuB, line 17: Can you remove this? It's not needed as the repo uses [melos](https://melos.invertase.dev)

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
```diff
@@ -464,22 +463,42 @@ private void finishWithSuccess(Object data) {
...
     }
 
-    private String getCauseFromException(Exception ex) {
+    private Map<String, Object> createErrorMap(@Nullable Exception cause) {
+        @Nullable AuthorizationException authException = cause instanceof AuthorizationException ? (AuthorizationException) cause : null;
+        Map<String, Object> responseMap = new HashMap<>();
+        responseMap.put("legacy_error_details", getCauseFromException(cause));
+
+        if (authException != null) {
+            responseMap.put("type", authException.type);
```
MaikuB, line 487: An exception is being thrown as on the Dart side `type` is defined as `String?` type but the AppAuth Android SDK has `type` defined as `int` so there's a type mismatch

flutter_appauth/example/lib/main.dart
SamCosta1: Yes 🤦 . Thought I'd removed that since making a PR, it was the easiest way to check  the errors were coming through properly

flutter_appauth/example/pubspec.yaml
SamCosta1: Ah right okay, is this something that's common knowledge in flutter (I've been using dart/flutter for a very short time) or is it worth adding this to documentation?

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
MaikuB: `code` seems to have a similar issue and possibly other scenarios too. Can you check Android overall?

flutter_appauth/example/pubspec.yaml
MaikuB: When it comes to Flutter overall, I'd say no. However, those who have been working on plugins/packages with monorepos for a while would likely know. The repo doesn't document so agreed there could be documentation added (e.g. in the contributing guide) but would be out of scope of the PR

SamCosta1: Agreed! :)

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
SamCosta1: Yeah I just noticed code as well, this does work so there must be some implicit type conversion going on somewhere which makes me uneasy. I'll go through and make sure they're being explicitly converted

flutter_appauth_platform_interface/lib/src/errors.dart
```diff
@@ -0,0 +1,82 @@
+import 'package:flutter/services.dart';
+
+class FlutterAppAuthPlatformErrorDetails {
```
MaikuB, line 3: Could API docs be added for the new classes and their properties?

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
MaikuB: I forgot to add the link but you can see the docs for the AppAuth Android SDK when it comes to the `AuthorizationException` at https://openid.github.io/AppAuth-Android/docs/latest/net/openid/appauth/AuthorizationException.html

MaikuB: > Yeah I just noticed code as well, this does work so there must be some implicit type conversion going on somewhere which makes me uneasy

I'm not seeing myself. When running the example app,  a `TypeError` being passed to the `clearBusyState()` method


image

SamCosta1: That's strange, it works for me but I'm suddenly getting a type error for the user cancellation boolean, despite it all working after my initial tests... Can't think what could have changed, I'll make sure it's bulletproof anyway

MaikuB:
> Yes exactly, the logout use case was the driver for creating this PR

@SamCosta1 you about to elaborate on the use case? Curious what it is so I can try to reproduce it

SamCosta1:
> > Yes exactly, the logout use case was the driver for creating this PR
> 
> @SamCosta1 you about to elaborate on the use case? Curious what it is so I can try to reproduce it

Of course! :) 

On the apps I've been working on the way we determine if a user is authenticated is by checking if we have tokens cached to disk. No tokens = not logged in.

When we use a refresh token to refresh an access token, this can obviously fail. One reason for this is because the refresh token has expired. In this case we must clear the locally cached tokens and log the user out (otherwise the user is in an irrecoverable state). 

Without this PR it's impossible to know _why_ a token refresh failed so we're forced to perform this logout on any failure. (This isn't quite true, we can detect a failure due to network issues but still). We only want to log the user out in certain cases e.g `invalid_grant` when the state is irrecoverable.

Without these granular checks, downtime in the authorization server could cause mass logouts. For example if at peak time the authorization server starts returning 500 errors, many users could be logged out at once which adds system load as they try to get back in, and more importantly is terrible user experience.

flutter_appauth_platform_interface/lib/src/errors.dart
```diff
@@ -0,0 +1,112 @@
+import 'package:flutter/services.dart';
+
+/// The details of an error thrown from the underlying
+/// platform AppAuth libraries.
```
MaikuB, line 4: IMO the sentence doesn't make sense and/or reads awkward, particularly around "underlying platform AppAuth libraries". "Underlying platform" results in a single platform (Android or iOS) so there should only be one library. 

Something along the lines of


```
/// The details of an error thrown from the underlying
/// platform's AppAuth SDK
```

would IMO make more sense. It also uses "SDK" that aligns with how AppAuth for Android and iOS refers to themselves as SDKs so brings consistency there

flutter_appauth_platform_interface/lib/src/errors.dart
```diff
@@ -0,0 +1,112 @@
...
+      error: map['error'],
+      errorDescription: map['error_description'],
+      errorUri: map['error_uri'],
+      rootCauseError: map['root_cause_error'],
+      userDidCancel: map['user_did_cancel']?.toLowerCase().trim() == 'true',
+    );
+  }
+}
+
+/// Exception that can be thrown by methods that launch a browser session
```
MaikuB, line 69: Could you look at shortening the sentence or breaking it up?

flutter_appauth_platform_interface/lib/src/errors.dart
```diff
@@ -0,0 +1,112 @@
...
+    details: legacyDetails,
+    stacktrace: stacktrace,
+  );
+
+  @override
+  String toString() {
+    return 'FlutterAppAuthUserCancelledException{platformErrorDetails: $platformErrorDetails}';
+  }
+
+  final FlutterAppAuthPlatformErrorDetails platformErrorDetails;
```
MaikuB, line 91: Bit of inconsistency here as properties were defined before methods

flutter_appauth_platform_interface/lib/src/errors.dart
MaikuB: On a related note, if you agree with this, could you update other similar API docs in a similar way?

flutter_appauth_platform_interface/lib/src/errors.dart
```diff
@@ -0,0 +1,112 @@
...
+
+  /// For 400 errors from the Authorization server, this is error defined here: https://datatracker.ietf.org/doc/html/rfc6749#section-5.2
+  /// e.g "invalid_grant"
+  /// Otherwise a short error describing what happened.
+  final String? error;
+
+  /// Short, human readible error description.
+  final String? errorDescription;
+
+  /// The error Uri or domain from the underlying platform.
```
MaikuB, line 39: Could you replace "Uri" with "URI"? It's an acronym so should be capitalised

flutter_appauth_platform_interface/lib/src/errors.dart
MaikuB: Thanks for making the changes. Continuing on this thread to leave a general comment. Could you look at https://dart.dev/effective-dart/documentation to see how the API docs could be updated to match. The ones that stand out to me so far around how some properties aren't  starting using a single sentence summary and could then be followed up with more information in separate paragraphs

flutter_appauth_platform_interface/lib/src/errors.dart
```diff
@@ -0,0 +1,112 @@
...
+  ///          https://github.com/openid/AppAuth-Android/blob/c6137b7db306d9c097c0d5763f3fb944cd0122d2/library/java/net/openid/appauth/AuthorizationException.java#L158
+  /// Recommend not using this field unless you really have to, see `error` field below.
+  final String? code;
+
+  /// For 400 errors from the Authorization server, this is error defined here: https://datatracker.ietf.org/doc/html/rfc6749#section-5.2
+  /// e.g "invalid_grant"
+  /// Otherwise a short error describing what happened.
+  final String? error;
+
+  /// Short, human readible error description.
```
MaikuB, line 36: typo: should be "readable"

flutter_appauth_platform_interface/lib/src/errors.dart
```diff
@@ -0,0 +1,112 @@
...
+  final String? errorDescription;
+
+  /// The error Uri or domain from the underlying platform.
+  final String? errorUri;
+
+  /// The underlying raw error as a String debugging.
+  final String? rootCauseError;
```
MaikuB, line 43: A few points on this one
- I don't understand what this means so I suspect others won't either
- The description doesn't line up with the name so something seems off here
- I took a look further on this and on iOS this references a `debugDescription`. Apple's own [docs](https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418703-debugdescription) say this is textual representation of the object itself for debugging purposes. This reads like the equivalent of calling `toString()` on the object itself and aligns with what I've seen the values I've seen.  On Android, this references a cause and this means the original error/exception. Based on this, these refer to different things so wouldn't be appropriate to map to the same property. I would suggest keeping this to Android only and mention as much in the API docs. Unless I'm missing something, the `debugDescription` from iOS seems redundant given it's a single string representation that consists of the code, domain and user info. Code and domain are already returned. If the user info dictionary is useful then I would suggest having a separate property that is documented as being specific to iOS/macOS

flutter_appauth/README.md
```diff
@@ -107,6 +119,25 @@ await appAuth.endSession(EndSessionRequest(
...
+try {
+  await appAuth.authorize(...);
+} on FlutterAppAuthPlatformException catch (e) {
+  final FlutterAppAuthPlatformErrorDetails details = e.details;
+  // Handle exceptions based on errors from AppAuth.
+} catch (e) {
+  // Handle other errors.
+}
+
+The `FlutterAppAuthPlatformErrorDetails` object contains all the error information from the underlying platforms, including what's speficied in the [RFC](https://datatracker.ietf.org/doc/html/rfc6749#section-5.2).
```
MaikuB, line 137: A few comments on this one
- Just noticed this having having left a similar comment below.  IMO, should replace "underlying platforms" with "underlying platform's AppAuth SDK". The name of the class already indicates it's for a single platform that using the plural form could create confusion
- The sentence is rather long too and there's opportunity to break up the part following the comma into its own sentence
- There's a typo "speficied" -> "specified"

flutter_appauth_platform_interface/lib/src/errors.dart
SamCosta1: I slightly disagree with this, though ultimately it's your call. (I've updated the comment though since it's misleading).

The underlying error (the `cause` field on Android and the error on iOS) are untyped properties so theoretically could contain other bits of information that I've missed. Especially in future changes to the underlying platform's AppAuth SDKs which might be easy to miss while upgrading this library.

Having a catch all property that ensures everything is exposed to dart code means that developers are always guaranteed to have all the available information in some form, if nothing else it makes it easier to then raise a PR on this library to add additional properties.

MaikuB: I have a suspicion what I said has been misunderstood so will try to elaborate. Based on your response, I would guess that you think I'm saying to not have this property or to remove information. If so, that isn't what I'm saying. I understand your intent to have a catch all property and have no issue with providing more information. My main point is what this represents on each platform is different. For iOS/macOS, this represents the error itself. It's only on Android that this represents what **caused** of the error. This means it represents another error entirely that the name of `rootCauseError` only make sense for Android. It also explains why on Android that when you close the browser that the `rootCauseError` is an empty string. This is because it doesn't contain the details of the `AuthorizationException` itself, which would've made it consistent with iOS/macOS. Rather, it contains the details of another error/exception that caused the `AuthorizationException` to be thrown.

Hope this is clearer and why I mentioned having two separate properties (one for iOS and one for Android) makes more sense.

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
```diff
@@ -464,22 +463,43 @@ private void finishWithSuccess(Object data) {
...
+        responseMap.put("legacy_error_details", getCauseFromException(cause));
+
+        if (authException != null) {
+            boolean userDidCancel = authException.equals(AuthorizationException.GeneralErrors.USER_CANCELED_AUTH_FLOW);
+            responseMap.put("type", String.valueOf(authException.type));
+            responseMap.put("code", String.valueOf(authException.code));
+            responseMap.put("error", authException.error);
+            responseMap.put("error_description", authException.errorDescription);
+            responseMap.put("error_uri", authException.errorUri == null ? null : authException.errorUri.toString());
+            responseMap.put("root_cause_error", authException.getCause() == null ? "" : authException.getCause().toString());
```
MaikuB, line 493: Can this be made to return a null instead of an empty string when the cause is null?

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
```diff
@@ -464,22 +463,43 @@ private void finishWithSuccess(Object data) {
...
-        finishWithError(DISCOVERY_ERROR_CODE, String.format(DISCOVERY_ERROR_MESSAGE_FORMAT, ex.error, ex.errorDescription), getCauseFromException(ex));
+        finishWithError(DISCOVERY_ERROR_CODE, String.format(DISCOVERY_ERROR_MESSAGE_FORMAT, ex.error, ex.errorDescription), ex);
     }
 
     private void finishWithEndSessionError(AuthorizationException ex) {
-        finishWithError(END_SESSION_ERROR_CODE, String.format(END_SESSION_ERROR_MESSAGE_FORMAT, ex.error, ex.errorDescription), getCauseFromException(ex));
+        finishWithError(END_SESSION_ERROR_CODE, String.format(END_SESSION_ERROR_MESSAGE_FORMAT, ex.error, ex.errorDescription), ex);
+    }
+
+    private Map<String, Object> createErrorMap(@Nullable Exception cause) {
```
MaikuB, line 481: Just spotted this and may explain why there was a misunderstanding on what I wrote below. Prior to your PR, I was trying to extract extra information from the cause of the exception. The exception itself (note: not the cause) would most likely be an instance of an `AuthorizationException`. Your PR now changes it to look at the exception that's been thrown, not the cause. However, what's passed here isn't actually the cause of the exception but the exception itself. The name of the parameter is named `cause`. Perhaps this one of the factors that caused some confusion on which level of the exception/error hierarchy you're looking at

flutter_appauth_platform_interface/lib/src/errors.dart
MaikuB: Oops sorry looks like I made a mistake myself. Just realised again that on iOS you do look at the underlying error. Sorry for the confusion. Let me mark this as resolved as I spotted something else and will comment on the relevant part(s) of the PR

flutter_appauth/ios/Classes/FlutterAppAuth.m
```diff
@@ -37,8 +37,49 @@ + (NSMutableDictionary *)processResponses:(OIDTokenResponse*) tokenResponse auth
...
+        if (underlyingError) {
+            if (underlyingError.domain) {
+                [details setObject:underlyingError.domain forKey:@"error_uri"];
+            }
+
+            if (underlyingError.debugDescription) {
+                [details setObject:underlyingError.debugDescription forKey:@"root_cause_error"];
+            }
+        } else if (error.debugDescription) {
+            [details setObject:error.debugDescription forKey:@"root_cause_error"];
```
MaikuB, line 74: From what I could tell, this is a part that could result in confusion. If there's no cause, the text representation of the error is being mapped to a property that represents the cause that would be non-existent. On Android, it will only ever hold information relevant to the cause.

This makes me think in order to achieve what you were trying to do in capturing as much information as possible, the PR would be in need of two properties

1. `rootCauseError`: this what you already have but it should only have populated if the cause exists. On iOS/macOS, it comes from `underlyingError.debugDescription`. On Android, it captures the result of calling `toString()` on the cause of `AuthorizationException`
2. A new property. On iOS, it comes from the `debugDescription` of  the `error` variable. Android, it represents calling `toString()` on the `AuthorizationException`. I can't think of a name for this property 

The intent here is to ensure the information consistently comes from the same level and the name of the property on the Dart side reflects this. Does that make sense?

flutter_appauth/ios/Classes/FlutterAppAuth.m
```diff
@@ -37,8 +37,49 @@ + (NSMutableDictionary *)processResponses:(OIDTokenResponse*) tokenResponse auth
...
+        }
+        if (error.code) {
+            [details setObject:[@(error.code) stringValue] forKey:@"code"];
+        }
+        
+        id underlyingErr = [error.userInfo objectForKey:NSUnderlyingErrorKey];
+        NSError *underlyingError = [underlyingErr isKindOfClass:[NSError class]] ? underlyingErr : nil;
+        if (underlyingError) {
+            if (underlyingError.domain) {
+                [details setObject:underlyingError.domain forKey:@"error_uri"];
```
MaikuB, line 67: If there's an intent to map the `domain` then unless I'm missing something, this should go to a separate property instead of here. The reason I mention this is that OAuth specs are specific on what  `error_uri` means (https://datatracker.ietf.org/doc/html/rfc6749#section-5.2). The SDKs extract this from reading the `error_uri` query string parameter of the redirect response.  On Android, this goes to a dedicated property so is handled more cleanly. From what I could tell, the AppAuth iOS SDK doesn't do so nor does it set it as the domain. [This code](https://github.com/openid/AppAuth-iOS/blob/c89ed571ae140f8eb1142735e6e23d7bb8c34cb2/Sources/AppAuthCore/OIDErrorUtilities.m#L104) indicates it's part of the description in the `NSError`'s `userInfo` dicationary

Based on this, if `errorUri` will be kept on the Dart side then it would be more accurate to tell devs that this is only populate on Android and that extracting it from iOS needs to be done differently. The latter may be difficult to do so I'm not expecting it to be covered in the PR. I believe for this to be done more easily, the AppAuth iOS SDK would need another key to save the value of the `error_uri` against the `userInfo`

SamCosta1: Ah yes you're right, I think I just assumed that was directly mapped 🤦 . I'll expose the domain an iOS as a separate property just so it's not missed and only map the error uri on Android

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
MaikuB: Realised there was a lot of text there but the summary/tl;dr is IMO better to rename `cause` to avoid confusion

flutter_appauth/ios/Classes/FlutterAppAuth.m
SamCosta1: Done :) I've split these into two properties

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
SamCosta1: Done :)

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
SamCosta1: Done :)

flutter_appauth/README.md
SamCosta1: Done :)

flutter_appauth_platform_interface/lib/src/errors.dart
SamCosta1: Done :)

flutter_appauth_platform_interface/lib/src/errors.dart
```diff
@@ -0,0 +1,146 @@
...
+
+  /// A debug description of the underlying cause of the error from the
+  /// platform's AppAuth SDK
+  final String? rootCauseDebugDescription;
+
+  /// Whether or not this error is caused by user cancellation
+  ///
+  /// Will only be `true` if the user cancelled the authorization flow by
+  /// closing the browser prematurely.
+  final bool userDidCancel;
```
MaikuB, line 75: Wanted to double check something I noticed whilst updating docs etc. If I'm not mistaken, this is redundant as there's an exception specific for catching a cancellation. Can you confirm?

SamCosta1: Technically yes you're right, but I did this intentionally since I can imagine cases of users wanting to pass the `FlutterAppAuthPlatformErrorDetails` object into a method responsible for handling all the various cases, it seemed cleaner to encapsulate everything on one object.

I don't have particularly strong feelings about this one though.

MaikuB: Thanks I've updated PR to remove this directly so I can merge this sooner. Will do a prerelease after though am looking to make it be part of a major release. The reason for this is I'm looking to bump the minimum Flutter SDK requirement to help make it easier to maintain. Would also allow for any potential breaking changes around error codes that have been discussed before

PR #523, opened 2024-08-10
[various] added FlutterAppAuthOAuthError to hold standard error codes that apps can reference

MaikuB:


SamCosta1:
@MaikuB I think this type still needs exporting? I can't access it from an app consuming the library (I may be wrong though).

MaikuB:
No you're right, I forgot to do so. Will fix this soon

MaikuB:
Should be fixed as part of 7.0.0-dev.4. Does this mean you are using the prerelease then? If so, let me know how it goes as I'm on making this a stable release. Note I ended up exposing these as strings to make the experience of using these easier and is the more direct approach. Was thinking that if an identity provider provided other values that having an enhanced enum or a class would make the developer experience bit worse. Have an app right now that could get `access_denied` as the error and isn't part of the RFC/spec

SamCosta1:
Yeah I'm using the pre-release version, no issues so far! I think the ideas discussed on the other thread to have much more robust cross platform error handling is the way to go though even if that means an enhanced enum. I'm needing to detect the case where the error's caused by lack of network connection and it's non trivial. It'd be great if this was just exposed directly as an error case.

SamCosta1:
Also thanks for removing the unneeded nullability, you beat me to it I was planning on raising a PR! 😅

PR #553, opened 2024-10-11
Add try catch to uncaught exception

john-slow:
As promised, I have created this pull request to address the bug where manually closing the login popup sheet could potentially trigger the following exception:
```
*** Terminating app due to uncaught exception 'An OAuth redirect was sent to a OIDExternalUserAgentSession after it already completed.', reason: 'An OAuth redirect was sent to a OIDExternalUserAgentSession after it already completed.'
*** First throw call stack:
(0x19f8a908c 0x19cbab2e4 0x19f9a4648 0x1027fe284 0x1037bf67c 0x10655f1a4 0x10652fe54 0x1031c95b0 0x1a3123f1c 0x1a312396c 0x1c2c5c248 0x1c2c2c68c 0x19f8460f4 0x19f845144 0x1038e671c 0x1038ea13c 0x1b8d5d3b8 0x1b8d5d338 0x1b8d5d210 0x19f87c088 0x19f87c01c 0x19f879b08 0x19f878d04 0x19f8785b8 0x1eb30e1c4 0x1a23ce2c0 0x1a247cddc 0x1a27a9b20 0x1032c47e8 0x1032c4760 0x1032c4864 0x1c504cd34)
libc++abi: terminating due to uncaught exception of type NSException
```

### DIFF
--- a/flutter_appauth/ios/Classes/FlutterAppauthPlugin.m
+++ b/flutter_appauth/ios/Classes/FlutterAppauthPlugin.m
@@ -483,12 +483,15 @@ AppAuthAuthorization *authorization;
             openURL:(NSURL *)url
             options:
                 (NSDictionary<UIApplicationOpenURLOptionsKey, id> *)options {
-  if ([_currentAuthorizationFlow resumeExternalUserAgentFlowWithURL:url]) {
-    _currentAuthorizationFlow = nil;
-    return YES;
-  }
-
-  return NO;
+    @try {
+      if ([_currentAuthorizationFlow resumeExternalUserAgentFlowWithURL:url]) {
+        _currentAuthorizationFlow = nil;
+        return YES;
+      }
+      return NO;
+    } @catch (NSException *exception) {
+        NSLog(@"Exception caught: %@, Reason: %@", exception.name, exception.reason);
+    }
 }
 
 - (BOOL)application:(UIApplication *)application
--- a/flutter_appauth/ios/Classes/FlutterAppauthPlugin.m
+++ b/flutter_appauth/ios/Classes/FlutterAppauthPlugin.m
@@ -483,15 +483,16 @@ AppAuthAuthorization *authorization;
             openURL:(NSURL *)url
             options:
                 (NSDictionary<UIApplicationOpenURLOptionsKey, id> *)options {
-    @try {
-      if ([_currentAuthorizationFlow resumeExternalUserAgentFlowWithURL:url]) {
-        _currentAuthorizationFlow = nil;
-        return YES;
-      }
-      return NO;
-    } @catch (NSException *exception) {
-        NSLog(@"Exception caught: %@, Reason: %@", exception.name, exception.reason);
+  @try {
+    if ([_currentAuthorizationFlow resumeExternalUserAgentFlowWithURL:url]) {
+      _currentAuthorizationFlow = nil;
+      return YES;
     }
+    return NO;
+  } @catch (NSException *exception) {
+    NSLog(@"Exception caught: %@, Reason: %@", exception.name,
+          exception.reason);
+  }
 }
 
 - (BOOL)application:(UIApplication *)application


flutter_appauth/ios/Classes/FlutterAppauthPlugin.m
```diff
@@ -483,12 +483,16 @@ - (BOOL)application:(UIApplication *)application
             openURL:(NSURL *)url
             options:
                 (NSDictionary<UIApplicationOpenURLOptionsKey, id> *)options {
-  if ([_currentAuthorizationFlow resumeExternalUserAgentFlowWithURL:url]) {
-    _currentAuthorizationFlow = nil;
-    return YES;
+  @try {
```
MaikuB, line 486: I'm having trouble getting the timing right to reproduce this but based on the code changes, from the perspective of an app using the plugin from the Dart/Flutter side, the exception is silently caught and won't be informed that there was a problem. Is my understanding correct? If it is, can the PR make the appropriate changes that would raise an exception for apps to handle? The other thing is based on what you mentioned, I suspect macOS would also have a similar problem

john-slow: This bug is somewhat rare, occurring about 1 to 2 times out of 20. I'd prefer to ignore this exception if possible, but the app crashes if it's not handled in objective-c. It works fine even if we don't raise the exception to Dart/Flutter, as long as it's managed in Objective-C. For macOS, I believe it would face the same issue if it uses the same login mechanism.

john-slow: I added a simple throw in the catch functions and also added try catch to macOS as well. 
Please let me know if you have any suggestion or question.

MaikuB: > Edit: Do you have a suggestion for an exception type to raise?

The plugin will handle throwing the appropriate exception provided you make the calls on the native method channel side. There are existing examples do that this on iOS and macOS by calling [this method](https://github.com/MaikuB/flutter_appauth/blob/1ceea0e7a322405d3749395f4ed4aa8620330daa/flutter_appauth/ios/Classes/FlutterAppAuth.m#L53). Calling this will raise the appropriate exception

I understand this is rare and how the PR in its current form avoids a crashed but since it's silently caught from the perspective of the Flutter side, it doesn't provide the app/developer the opportunity to catch the exception and leave it them to do decide what to do. If I'm not mistaken, the PR as it is would prevent an uncaught exception from going to something like Crashlytics too

john-slow: As you mentioned, the exception is typically handled in the code you provided. However, this specific error originates from the code [here](https://github.com/MaikuB/flutter_appauth/blob/1ceea0e7a322405d3749395f4ed4aa8620330daa/flutter_appauth/ios/Classes/FlutterAppauthPlugin.m#L481C1-L492C2). I've attached a video that should clarify things. In the video, I simply click the "done" button during the auto login process (no password entry required). There's a slight chance the app might either crash or freeze (in the video, it froze), and you can see the exception printed on the left side of the console.

https://github.com/user-attachments/assets/8366323b-4734-478e-8078-72a8fbaa6893

You're right in your understanding, but I'm uncertain about how can I help debug when this happens. My code is intended to prevent the app from crashing only. Could you either demonstrate or guide me on the proper way to handle this?

MaikuB: To clarify, I wasn't asking for helping to debug. I was asking for the code to be changed so that when the scenario happens, rather than silently swallowing the error, it results in an except that can by caught by applications. I understand this is rare but this would enable applications to know something happened and decide what to do. In terms of guidance, refer to the links in my previous post as the intent there to provide references to existing that shows how it's currently done to help guide. It has links to code that shows how the plugin raises exceptions starting from the native side and how that makes its way to the Flutter side

john-slow: OK, I see, thanks for your clarification.

deandreamatias:
I still with this issue when used flutter_appauth. Can take a look on this PR again @MaikuB and @john-slow ?
Thanks!

PR #585, opened 2025-01-10
[flutter_appauth] remove idTokenHint assert

clintpurser:
Hi @MaikuB 

I am wondering if there is a specific reason for this assert?

Currently we want to only supply our postLogoutRedirectUrl but not the idTokenHint for logging out in our app. We tried using the additionalParameters field to supply only the post_logout_redirect_uri, but that causes fatal exception in android.

Would be nice to remove this assert, unless there is a reason with OpenID why we'd want to enforce it like this.

Let me know!
Thank you!

### DIFF
--- a/flutter_appauth_platform_interface/lib/src/end_session_request.dart
+++ b/flutter_appauth_platform_interface/lib/src/end_session_request.dart
@@ -13,8 +13,7 @@ class EndSessionRequest with AcceptedAuthorizationServiceConfigurationDetails {
     String? issuer,
     String? discoveryUrl,
     AuthorizationServiceConfiguration? serviceConfiguration,
-  }) : assert((idTokenHint == null && postLogoutRedirectUrl == null) ||
-            (idTokenHint != null && postLogoutRedirectUrl != null)) {
+  }) {
     this.serviceConfiguration = serviceConfiguration;
     this.issuer = issuer;
     this.discoveryUrl = discoveryUrl;


MaikuB:
Hi @clintpurser, it's been a while so I don't remember the *exact* reasoning. I believe the goal was encouraging developers to providing all details or nothing at all. Typically both are given as giving both is indicates the app is providing details on who should be logged out and where the IdP should go to upon a successful logout. These are assertions though so shouldn't be impacting release builds

victoreronmosele:
Hi @MaikuB, I believe this PR should be merged in. 

While the OpenID spec recommends passing the `id_token_hint` together with the `post_logout_redirect_uri`, the spec also mentions an alternative where you pass the `client_id` together with the `post_logout_redirect_uri` when `id_token_hint` is omitted. 

> client_id
> OPTIONAL. ...The most common use case for this parameter is to specify the Client Identifier when post_logout_redirect_uri is used but id_token_hint is not...
>
> Source: [OIDC RP-Initiated Logout](https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout )

## Use Case: 

My use case is an app where the `id_token_hint` is very large due to a number of claims (orgs/roles/permissions) and causes the browser redirection to fail with `414 Request-URI Too Large` error when `id_token_hint` is passed during logout. 

The alternative is to omit the `id_token_hint` and instead, pass the `client_id` inside `additionalParameters`.

## Workaround

The current workaround is to create a custom that implements `EndSessionRequest`.

This bypasses the assertions and replaces `EndSessionRequest` in the `FlutterAppAuth.endSession` method.

<details><summary>Open to view custom class</summary>

```dart
class CustomEndSessionRequest implements EndSessionRequest {
  @override
  final String? idTokenHint;
  @override
  final String? postLogoutRedirectUrl;
  @override
  final String? state;
  @override
  @override
  ExternalUserAgent? externalUserAgent;
  @override
  final Map<String, String>? additionalParameters;
  @override
  AuthorizationServiceConfiguration? serviceConfiguration;
  @override
  String? issuer;
  @override
  String? discoveryUrl;

  CustomEndSessionRequest({
    this.idTokenHint,
    this.postLogoutRedirectUrl,
    this.state,
    this.allowInsecureConnections = false,
    this.externalUserAgent = ExternalUserAgent.asWebAuthenticationSession,
    this.additionalParameters,
    this.serviceConfiguration,
    this.issuer,
    this.discoveryUrl,
  });

  @override
  void assertConfigurationInfo() {
    assert(
      issuer != null || discoveryUrl != null || serviceConfiguration != null,
      'Either the issuer, discovery URL or service configuration must be '
      'provided',
    );
  }
}
```
 
**Usage:** 

```dart
final endSessionRequest = CustomEndSessionRequest(
  // Other properties...
  postLogoutRedirectUrl: postLogoutRedirectUrl,
  additionalParameters: {'client_id': clientId},
);
```

</details>

MaikuB:
Thanks @victoreronmosele for the information. Makes sense and I'll merge this in

PR #613, opened 2025-05-13
implement support for external custom browsers

### LINKED ISSUE #565: [IOS] Would it be possible to support non-embedded browser (so use the default system browser)
I have been using a fork of this library that adds the use of a default browser to authenticate. I had to use this becasue the way my authentication works is as follows: 

- user clicks on login button -> default browser opens (not in-app)  -> browser opens authentication app (itsme) -> app opens callback url back in the browser -> browser closes and goes back to app. 

This used to work just fine until iOS 18. If I click the login button I just get "error: unable to open Safari."

So I tried using the latest version of flutter_appauth. But then the flow breaks. it looks something like this.

- user clicks on login button -> in-app browser opens -> inapp browser opens authentication app (itsme) -> app opens callback url in default browser (so it doesnt return to the app with the in-app browser open) -> callback show an error.

CaringDev:
Adds support for external (non-in-app) browsers on iOS to participate in SSO-sessions and allow e.g. Citrix MDM to provide its secure browser (which in turn allows SSO for e.g. enterprise Kerberos environments).

Fixes #565 but probably raises the same questions that were left unanswered in https://github.com/MaikuB/flutter_appauth/pull/318
Given that this unblocks some users, it might still be valuable.

### DIFF
--- a/flutter_appauth/example/lib/main.dart
+++ b/flutter_appauth/example/lib/main.dart
@@ -126,6 +126,19 @@ class _MyAppState extends State<MyApp> {
                               ExternalUserAgent.sfSafariViewController),
                     ),
                   ),
+                if (Platform.isIOS)
+                  Padding(
+                    padding: const EdgeInsets.all(8.0),
+                    child: ElevatedButton(
+                      child: const Text(
+                        'Auto code exchange using system browser (iOS only)',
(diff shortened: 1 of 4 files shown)


jipspaul:
Any update concerning this option on main branch ? I still use the patch in production

PR #620, opened 2025-08-02
[flutter_appauth] feat: web support

eirikb:
Add web support with PKCE auth.
Only supports redirect! No popup (yet).

Ok, so I've been using `flutter_appauth` for a while, with great success. For web support I have implemented my own solution. Now I started yet another Flutter project where I add `flutter_appauth` (for mobile) and start copying over my auth solution (for web), but, this time I figured I could try to add the solution directly into `flutter_appauth`.

I'm aware of #70 and honestly don't know if this PR will solve that.  
It only works with redirect, and the "Sign in with auto code exchange" is the only button in example that will work.

### DIFF
--- a/flutter_appauth/example/lib/main.dart
+++ b/flutter_appauth/example/lib/main.dart
@@ -2,6 +2,7 @@ import 'dart:convert';
 import 'dart:io' show Platform;
 import 'dart:math';
 
+import 'package:flutter/foundation.dart' show kIsWeb;
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter_appauth/flutter_appauth.dart';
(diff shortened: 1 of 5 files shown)


MaikuB:
Sorry, somehow missed this one.

> I'm aware of https://github.com/MaikuB/flutter_appauth/issues/70 and honestly don't know if this PR will solve that.

No, this will not solve it. Whilst I have not looked at that thread in a very long time, I believe one of the key points raised back then and at least in the OP is how this plugin is a wrapper/bridge for AppAuth SDKs. It's part of the name of this plugin as a result and the name AppAuth carries a connotation of referring the SDKs mentioned at https://appauth.io/. Since the PR here doesn't rely on the SDKs then it does not meet the criteria of being a wrapper.

With that in mind, I'll be closing this and I'll also look to create a FAQ section around this.

eirikb:
@MaikuB Understandable. If I create a proper and working PR using https://github.com/openid/AppAuth-JS (oficcial appauth.io SDK) it would be accepted?

MaikuB:
@eirikb that is correct

PR #638, opened 2026-01-15
[flutter_appauth] fix: change presentingViewController to topMost viewController

JLavat:
AppAuth Flutter plugin is currently using rootViewController as presentingViewController when it needs to present a ViewController (SafariVC) for authentication.

If there is some additional ViewController being presented on top of the root FlutterViewController, then the presentation will fail. In this case, using rootViewController will return the root, i.e. FlutterViewController, whose view is NOT in the window hierarchy because there is another viewController presented on top of it, then the system will raise the following error:

Attempt to present UIViewController on UIViewController whose view is not in the window hierarchy.

To avoid this issue, at any time, when needing to present a ViewController, we should instead retrieve and use the topMost viewController.

### DIFF
--- a/flutter_appauth_platform_interface/pubspec.yaml
+++ b/flutter_appauth_platform_interface/pubspec.yaml
@@ -1,7 +1,7 @@
 name: flutter_appauth_platform_interface
 description: A common platform interface for the flutter_appauth plugin.
-version: 11.0.0
-homepage: https://github.com/MaikuB/flutter_appauth/tree/master/flutter_appauth_platform_interface
+version: 11.1.0
+homepage: https://github.com/plx-eva/flutter_appauth/tree/master/flutter_appauth_platform_interface
 
 environment:
   sdk: ^3.7.0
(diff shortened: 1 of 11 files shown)


MaikuB:
Additionally, I'm wondering if the PR should be changing to follow what was done in the `url_launcher` plugin [here](https://github.com/flutter/packages/pull/10549)

JLavat:
Hi,
This scenario happens when some UIViewController is presented on top of the FlutterViewController.
While this is not so common, it can happen for apps that mix flutter and native code/UI.
What was done in the url_launcher plugin seems better than the existing code, but it won't fix the scenario I described.
If you don't feel necessary to support this use case, I can close the PR, I will continue using my fork.

flutter_appauth_platform_interface/pubspec.yaml
```diff
@@ -1,7 +1,7 @@
 name: flutter_appauth_platform_interface
 description: A common platform interface for the flutter_appauth plugin.
-version: 11.0.0
-homepage: https://github.com/MaikuB/flutter_appauth/tree/master/flutter_appauth_platform_interface
+version: 11.1.0
+homepage: https://github.com/plx-eva/flutter_appauth/tree/master/flutter_appauth_platform_interface
```
mofogasy, line 4: Might be a mistake to update the url

PR #645, opened 2026-03-04
 [flutter_appauth] add cancellation method for as web authentication session

### LINKED ISSUE #643: [iOS] Expose API to cancel/clear pending ASWebAuthenticationSession on app launch
### Problem

On iOS, when using `ASWebAuthenticationSession` (via `authorizeAndExchangeCode` with externalUserAgent: `asWebAuthenticationSession` or `ephemeralAsWebAuthenticationSession`), if the user leaves the auth flow in a half-finished state (e.g. the system consent dialog or the in-app browser is shown, then the app is backgrounded or killed without the user tapping Continue/Cancel), iOS can keep that session in a “pending” state.

When the user opens the app again (cold start), the system re-presents that pending session: the “\[App Name\] Wants to Use \"[domain]\" to Sign In” consent dialog appears immediately on launch, without the user tapping Sign In. This is confusing and looks like the app is starting auth on its own.

We use ephemeral sessions for all user-initiated sign-in (`ephemeralAsWebAuthenticationSession`), and in that flow the consent dialog does not appear. The unwanted dialog only appears on app launch, which strongly suggests it is a resumed/pending session from a previous run, not a new session started by our code.

### Desired behaviour

A way to cancel or clear any pending ASWebAuthenticationSession at app startup so that:
- No stale consent dialog is shown when the app is opened.
- The user only sees auth UI when they explicitly tap Sign In (at which point we start a new, ephemeral session as today).

### Proposed solution

Expose a Dart API that forwards to the native layer to cancel any active/pending session, for example:

- Option A: `Future<void> cancelPendingSession()` (or similar) on `FlutterAppAuth`, implemented on iOS by calling `cancel()` on any `ASWebAuthenticationSession` instance the plugin or AppAuth-iOS still holds. - [ASWebAuthenticationSession docs](https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession)
- Option B: Document that the plugin does not hold a reference across app launches, and if that’s the case, document the limitation and any recommended workaround (e.g. user must complete or cancel the flow).

You would call this method early in app startup (e.g. after plugin/GetIt init, before showing the main UI) so that any session left pending from a previous run is cleared and not re-presented. 

This has cause us problems when using conjunction with LocalAuth. As the biometric prompt and the Auth Dialogue start looping.

### Environment

-Package version: flutter_appauth ^10.0.0
- Platform: iOS (real device and simulator)
- Flow: OAuth 2 / OIDC with B2C-style provider; redirect URI `msauth.[bundleId]://auth`

### Notes

We are not asking to change when the consent is shown for a new session (that’s an iOS behaviour). We only need a way to clear a previous session that is being re-presented on launch.

If the plugin or AppAuth-iOS does not retain a reference to the session across process restarts, we’d still find it helpful if that were documented, and we’re happy to contribute a PR if a cancel API is feasible on the native side.

kodejack:
#643

### DIFF
--- a/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
+++ b/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
@@ -52,6 +52,7 @@ public class FlutterAppauthPlugin
   private static final String AUTHORIZE_METHOD = "authorize";
   private static final String TOKEN_METHOD = "token";
   private static final String END_SESSION_METHOD = "endSession";
+  private static final String CANCEL_PENDING_SESSION_METHOD = "cancelPendingSession";
 
   private static final String DISCOVERY_ERROR_CODE = "discovery_failed";
   private static final String AUTHORIZE_AND_EXCHANGE_CODE_ERROR_CODE =
@@ -197,6 +198,10 @@ public class FlutterAppauthPlugin
         } catch (Exception ex) {
           finishWithError(END_SESSION_ERROR_CODE, ex.getLocalizedMessage(), ex);
         }
+        break;
+      case CANCEL_PENDING_SESSION_METHOD:
+        // No-op on Android; only iOS/macOS use ASWebAuthenticationSession.
+        result.success(new HashMap<>());
         break;
       default:
         result.notImplemented();
--- a/flutter_appauth/macos/flutter_appauth/Sources/flutter_appauth/FlutterAppauthPlugin.m
+++ b/flutter_appauth/macos/flutter_appauth/Sources/flutter_appauth/FlutterAppauthPlugin.m
@@ -163,6 +163,11 @@ AppAuthAuthorization *authorization;
     [self handleTokenMethodCall:[call arguments] result:result];
   } else if ([END_SESSION_METHOD isEqualToString:call.method]) {
     [self handleEndSessionMethodCall:[call arguments] result:result];
+  } else if ([CANCEL_PENDING_SESSION_METHOD isEqualToString:call.method]) {
+    [authorization cancelPendingSessionWithCompletion:^{
+      self.currentAuthorizationFlow = nil;
+      result(@{});
+    }];
   } else {
     result(FlutterMethodNotImplemented);
   }
--- a/flutter_appauth/README.md
+++ b/flutter_appauth/README.md
@@ -267,6 +267,10 @@ API docs can be found [here](https://pub.dartlang.org/documentation/flutter_appa
 
 ## FAQs
 
+**On iOS/macOS, the "Wants to Use [domain] to Sign In" consent dialog appears automatically when I open the app (e.g. after the app was backgrounded during login). How do I clear it?**
+
+Call [cancelPendingSession](https://pub.dev/documentation/flutter_appauth/latest/flutter_appauth/FlutterAppAuth/cancelPendingSession.html) early at app startup (e.g. after plugin init, before showing the main UI). This cancels any pending [ASWebAuthenticationSession](https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession) so the system does not re-present it on launch. See [issue #643](https://github.com/MaikuB/flutter_appauth/issues/643). Note: If the app process was killed (cold start), the plugin does not retain a reference to the session; this API helps when the app was only backgrounded.
+
 **When connecting to Azure B2C or Azure AD, the login request redirects properly on Android but not on iOS. What's going on?**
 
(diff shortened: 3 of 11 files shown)


MaikuB:
Thanks for the PR. Can you add some details to your original post on what this PR solves. Currently I'm taking a guess from the API docs for the new method. I would be keen to see if you're able to share info on how to reproduce the problem and/or a recording. Is the example app  impacted by the problem this that PR solves? If so, can the example app be updated to show how it makes use of the addition as well?

vkammerer:
We would be keen to see this PR merged as our application needs it.

Our use case is that we need to close the `ASWebAuthenticationSession` view from the flutter application before it has completed.

MaikuB:
@vkammerer any chance you show a video and/or demo code using what this PR provides? I didn't get what you mean by your explanation. My understanding is this is about closing the browser and that's normally handled as part of the redirect so wondering if I'm missing/misunderstanding something

vkammerer:
@MaikuB our use case is that we want to close the `ASWebAuthenticationSession` before it redirects normally. To be more precise, the flow is as follows:
- we open an `ASWebAuthenticationSession`
- the user is taken through a web flow that can sometimes finish by tapping an app link / universal link in the browser, or sometimes lead to opening these app links / universal links from third party applications
- in both cases, the flutter app will handle these links while the `ASWebAuthenticationSession` is still open
- then we want to close the `ASWebAuthenticationSession` and present our Flutter implementation of the flow's completion.

MaikuB:
~~Thanks @vkammerer for the info. At which point does the cancellation method get called and how does it get called? This was one of the reasons I wanted to see if the example app can be updated to show the usage. On that note, does the example app fit the criteria for using the method? So far I can tell the method can close the modal browser by making minor tweaks (e.g. don't await the call that initiates the await request and call the cancellation method after some delay). However, a redirection typically closes the browser so I'm wondering what I'm missing.~~

~~Note I'm aware it won't hurt to the method but trying to get more understanding, seeing if it's something suits being in the example app and if there are some examples of usages that I could add through API docs and/or the readme~~

~~Edit: is there something that makes this unique to iOS/macOS as well?~~

Sorry I forgot/missed that @kodejack linked this PR to #643

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
```diff
@@ -198,6 +199,10 @@ public void onMethodCall(MethodCall call, @NonNull Result result) {
           finishWithError(END_SESSION_ERROR_CODE, ex.getLocalizedMessage(), ex);
         }
         break;
+      case CANCEL_PENDING_SESSION_METHOD:
+        // No-op on Android; only iOS/macOS use ASWebAuthenticationSession.
+        result.success(new HashMap<>());
```
MaikuB, line 204: Was there a reason to pass an empty `HashMap<>()`? I believe passing null would suffice

flutter_appauth/macos/flutter_appauth/Sources/flutter_appauth/FlutterAppauthPlugin.m
```diff
@@ -163,6 +163,11 @@ - (void)handleMethodCall:(FlutterMethodCall *)call
     [self handleTokenMethodCall:[call arguments] result:result];
   } else if ([END_SESSION_METHOD isEqualToString:call.method]) {
     [self handleEndSessionMethodCall:[call arguments] result:result];
+  } else if ([CANCEL_PENDING_SESSION_METHOD isEqualToString:call.method]) {
+    [authorization cancelPendingSessionWithCompletion:^{
+      self.currentAuthorizationFlow = nil;
+      result(@{});
```
MaikuB, line 169: Similar to my comment on Android side, from what I've seen and tried, calling `result(nil)` would have sufficed

MaikuB (changes requested):
So sorry about confusion @kodejack. The original post still kept the PR message template that is meant to be replaced that I missed you had linked to the issue number underneath it to describe about the problem in more depth. Functionality-wise PR is fine but would you be able to look into the following

- rename the method. The use of the word "session" refers to `ASWebAuthenticationSession` but this ends up causing a clash with the notion of a session that refers to a session after a user has logged in. I would recommend something like `dismissExternalUserAgent` as it's close to the name used within the AppAuth iOS SDK and the notion of an external user agent already exists in this plugin. The documentation would end needing adjustment as a result
- The [API docs](https://openid.github.io/AppAuth-iOS/docs/latest/protocol_o_i_d_external_user_agent-p.html#a83d57c32ab8aedce339a2655a2e4695c) for the method within the AppAuth iOS SDK takes a boolean parameter to indicate if the dismissal will be animated or note. Would you be able to expose a similar boolean parameter?

PR #647, opened 2026-03-12
Add autoGenerateState flag to EndSessionRequest to control whether to auto-generate state parameter

### LINKED ISSUE #646: Allow explicitly null state in EndSessionRequest
The OpenID Connect RP-Initiated Logout specification defines the state parameter as OPTIONAL. However, the current Android implementation of `flutter_appauth` makes it effectively mandatory due to the underlying behavior of the plugin's bridge logic with the AppAuth-Android and iOS.

# The Problem:
When calling `endSession` with an IdP that does not support/return the state parameter, the logout flow fails with a exception:
```dart
PlatformException(end_session_failed, Failed to end session: [error: null, description: Response state param did not match request state] ...)
```

##  Technical Root Cause:
   1. In the AppAuth-Android SDK, the EndSessionRequest.Builder constructor automatically generates a random, secure state string by default.
   2. In the `flutter_appauth` Android plugin, the `state` from Dart is only passed to the native builder if it is NOT null.
   3. Consequently, passing state `null` from Dart fails to clear the pre-generated state in the native builder. The request is sent with a state, the server returns without one, and AppAuth-Android throws an exception.

# Proposed Change:
Align the Android implementation with the OIDC specification by allowing the state to be explicitly set to null on the native code.

The `flutter_appauth`  logic should be updated to ensure that if `null` (or something equivalent, maybe a empty string '' or having an additional boolean flag) is passed from Dart, the native builder.setState(null) is called to override the default generated state.

**NOTE**: The AppAuth-Android SDK explicitly added support for a nullable `state` (see Issue [#615](https://github.com/openid/AppAuth-Android/issues/615)) to support IdP providers that do not echo the _state_ back in the logout redirect. This change would expose that native capability to Flutter developers.

pcba-dev:
## Problem
Some Identity Providers reject the end-session request because AppAuth (Android, iOS, macOS) automatically generates a random `state` value by  default. When those IdPs do not echo the state back, response validation fails and the logout flow breaks.

  | \`state\` | \`autoGenerateState\` | Behaviour |
  |---|---|---|
  | `null` | `true` (default) | AppAuth auto-generates _state_ — **unchanged** |
  | `"value"` | `true` | Explicit non-empty "value" used as _state_ — **unchanged** |
  | `"value"` | `false` | Explicit non-empty "value" used as _state_ — **unchanged** |
  | `null` | `false` | State suppressed — no _state_ sent to IdP → state suppressed |

---
Fixes #646

### DIFF
--- a/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
+++ b/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
@@ -287,6 +287,7 @@ public class FlutterAppauthPlugin
     final String idTokenHint = (String) arguments.get("idTokenHint");
     final String postLogoutRedirectUrl = (String) arguments.get("postLogoutRedirectUrl");
     final String state = (String) arguments.get("state");
+    final boolean autoGenerateState = (boolean) arguments.get("autoGenerateState");
     final boolean allowInsecureConnections = (boolean) arguments.get("allowInsecureConnections");
     final String issuer = (String) arguments.get("issuer");
     final String discoveryUrl = (String) arguments.get("discoveryUrl");
@@ -298,6 +299,7 @@ public class FlutterAppauthPlugin
         idTokenHint,
         postLogoutRedirectUrl,
         state,
+        autoGenerateState,
         issuer,
         discoveryUrl,
         allowInsecureConnections,
@@ -570,7 +572,7 @@ public class FlutterAppauthPlugin
           Uri.parse(endSessionRequestParameters.postLogoutRedirectUrl));
     }
 
-    if (endSessionRequestParameters.state != null) {
+    if (endSessionRequestParameters.state != null || !endSessionRequestParameters.autoGenerateState) {
       endSessionRequestBuilder.setState(endSessionRequestParameters.state);
     }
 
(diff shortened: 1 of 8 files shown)


MaikuB (changes requested):
Thanks for the PR. Presumably, the addition of the  `autoGenerateState` was done to minimise breaking changes. However, it looks it creates confusion in the semantics and requires a table like the one you provided to understand. For example,  when `autoGenerateState` is `true`, this will not actually mean an automatically generated state value is used if an explicit state value was given. IMO, the root cause here is because there's now two parameters that dictate the behaviour. The idea that comes to mind for me to solve this is to use inheritance like so

```
abstract class State {
  String? get value;
}

class NullState implements State {
  @override
  String? get value => null;
}

class RandomState implements State {

  @override
  String? get value => ...
}

class CustomState implements State {
  final String _value;

  CustomState(this._value);

  @override
  String? get value => _value;
}
```

There may be better names for these classes but this would be how these are used

- `NullState`: follows a null object pattern approach. The plugin interprets to nullify the default random state so the IdP gets a null state value
- `RandomState`: plugin ensures a random state value is used to then go to the IdP. The calculation of the random state value could end up being lifted up to be done by this plugin in Dart instead of the underlying SDKs as well
- `CustomState`: allows users of the plugin to explicitly provide a state value that is passed to the SDKs and then go to the IdP

This would require breaking changes but this would result in code that is easier to interpret/understand and could also be a pattern applied to other parts of the plugin (e.g. could be applied to the auth requests for state and nonce). Let me know your thoughts and if this something you could tackle in this PR

PR #650, opened 2026-04-26
Fix InsecureConnectionBuilder (Android) and add allowInsecureConnections support (iOS/macOS)

### LINKED ISSUE #386: PlatformException(authorize_and_exchange_code_failed, Failed to authorize: Connection error making token request to 'https://xxx/realms/AAA/protocol/openid-connect/token': The certificate for this server is invalid. You might be connecting to a server that is pretending to be 'xxx' which could put your confidential information at risk...
Hi to Flutter app auth team.

I'm using the last version of flutter_appauth to connect with Keycloak server. Everything are working well, but I got an error when callback in IOS(it's working fine on Android).
`PlatformException(authorize_and_exchange_code_failed, Failed to authorize: Connection error making token request to 'https://xxx/realms/AAA/protocol/openid-connect/token': The certificate for this server is invalid. You might be connecting to a server that is pretending to be 'xxx' which could put your confidential information at risk...`

For iOS in info.plist:
`<key>CFBundleURLTypes</key>
	<array>
		<dict>
			<key>CFBundleTypeRole</key>
			<string>Editor</string>
			<key>CFBundleURLSchemes</key>
			<array>
				<string><my_scheme></string>
			</array>
		</dict>
	</array>`
Maybe it's problem from self-signed certificate. How could I solve this problem?

jrevillard:
## Summary

Fixes #386

The `allowInsecureConnections` parameter does not work on any platform:

- **Android**: `InsecureConnectionBuilder.openConnection()` is a no-op — it calls `URL.openConnection()` without any SSL/TLS configuration, so HTTPS connections still validate the certificate chain normally.
- **iOS/macOS**: The `allowInsecureConnections` parameter is completely ignored. There is no equivalent of `InsecureConnectionBuilder`, and no SSL bypass mechanism exists.

This makes it impossible to use the plugin against OIDC providers with self-signed certificates during development (e.g., local Keycloak instances).

### Android
Replaced the no-op `InsecureConnectionBuilder.openConnection()` with a real implementation that:
- Creates an `SSLContext` with a trust-all `X509TrustManager`
- Sets a permissive `HostnameVerifier` on `HttpsURLConnection`
- Only applies when `allowInsecureConnections=true` (the existing routing in `FlutterAppauthPlugin` via `getConnectionBuilder()` is unchanged)

### iOS & macOS
Added `InsecureURLProtocol` (an `NSURLProtocol` subclass) that:
- Intercepts HTTPS requests when registered
- Creates a private `NSURLSession` with empty `protocolClasses` to avoid recursion
- Trusts all server certificates via `NSURLSessionAuthChallengeUseCredential`
- Is registered before and unregistered after each plugin method call, only when `allowInsecureConnections=true`

Also added `allowInsecureConnections` property to `TokenRequestParameters` and `EndSessionRequestParameters` on iOS/macOS (was previously only defined on Android).

## Testing

Verified on Android emulator against a local Keycloak instance with a self-signed certificate:
- Endpoint discovery
- Authorization + token exchange
- Token refresh
- End session (logout)

All operations succeed without "Network error" when `allowInsecureConnections=true`.

## Security Note

The SSL bypass is **only active** when `allowInsecureConnections=true` is explicitly passed by the caller. Production configurations use CA-signed certificates and `allowInsecureConnections=false` (the default), so this change has no production impact.

### DIFF
--- a/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/InsecureConnectionBuilder.java
+++ b/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/InsecureConnectionBuilder.java
@@ -9,6 +9,12 @@ import net.openid.appauth.connectivity.ConnectionBuilder;
 import java.io.IOException;
 import java.net.HttpURLConnection;
 import java.net.URL;
+import java.security.SecureRandom;
+import java.security.cert.X509Certificate;
+import javax.net.ssl.HttpsURLConnection;
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.X509TrustManager;
 
 public class InsecureConnectionBuilder implements ConnectionBuilder {
 
@@ -19,6 +25,21 @@ public class InsecureConnectionBuilder implements ConnectionBuilder {
   @NonNull
   @Override
   public HttpURLConnection openConnection(@NonNull Uri uri) throws IOException {
-    return (HttpURLConnection) new URL(uri.toString()).openConnection();
+    HttpURLConnection conn = (HttpURLConnection) new URL(uri.toString()).openConnection();
+    if (conn instanceof HttpsURLConnection) {
+      try {
+        SSLContext sslContext = SSLContext.getInstance("TLS");
+        sslContext.init(null, new TrustManager[] { new X509TrustManager() {
+          public void checkClientTrusted(X509Certificate[] chain, String authType) {}
+          public void checkServerTrusted(X509Certificate[] chain, String authType) {}
+          public X509Certificate[] getAcceptedIssuers() { return new X509Certificate[0]; }
+        }}, new SecureRandom());
+        ((HttpsURLConnection) conn).setSSLSocketFactory(sslContext.getSocketFactory());
+        ((HttpsURLConnection) conn).setHostnameVerifier((hostname, session) -> true);
+      } catch (Exception e) {
+        throw new IOException("Failed to configure insecure connection", e);
+      }
+    }
+    return conn;
   }
 }
--- a/flutter_appauth/ios/flutter_appauth/Sources/flutter_appauth/FlutterAppAuth.h
+++ b/flutter_appauth/ios/flutter_appauth/Sources/flutter_appauth/FlutterAppAuth.h
@@ -56,6 +56,7 @@ static NSString *const END_SESSION_ERROR_MESSAGE_FORMAT =
 @property(nonatomic, strong) NSDictionary *serviceConfigurationParameters;
 @property(nonatomic, strong) NSDictionary *additionalParameters;
 @property(nonatomic, strong) NSNumber *externalUserAgent;
+@property(nonatomic, assign) BOOL allowInsecureConnections;
 @end
 
 typedef NS_ENUM(NSInteger, ExternalUserAgent) {
--- a/flutter_appauth/ios/flutter_appauth/Sources/flutter_appauth/FlutterAppauthPlugin.m
+++ b/flutter_appauth/ios/flutter_appauth/Sources/flutter_appauth/FlutterAppauthPlugin.m
@@ -2,6 +2,90 @@
 
 #import "FlutterAppauthPlugin_Private.h"
 
+#pragma mark - InsecureURLProtocol
+
+@interface InsecureURLProtocol : NSURLProtocol <NSURLSessionDataDelegate>
+@property (nonatomic, strong) NSURLSessionDataTask *dataTask;
+@property (nonatomic, strong) NSURLSession *session;
+@end
+
+static NSString *const kInsecureURLProtocolHandledKey = @"InsecureURLProtocolHandled";
+
+@implementation InsecureURLProtocol
+
++ (BOOL)canInitWithRequest:(NSURLRequest *)request {
+  return [self propertyForKey:kInsecureURLProtocolHandledKey inRequest:request] == nil;
+}
+
++ (NSURLRequest *)canonicalRequestForRequest:(NSURLRequest *)request {
+  return request;
+}
+
+- (void)startLoading {
+  NSMutableURLRequest *mutableRequest = [self.request mutableCopy];
+  [InsecureURLProtocol setProperty:@YES
+                            forKey:kInsecureURLProtocolHandledKey
+                         inRequest:mutableRequest];
+
+  NSURLSessionConfiguration *config =
+      [NSURLSessionConfiguration defaultSessionConfiguration];
+  config.protocolClasses = @[];
+
+  self.session = [NSURLSession sessionWithConfiguration:config
+                                               delegate:self
+                                          delegateQueue:[NSOperationQueue mainQueue]];
+  self.dataTask = [self.session dataTaskWithRequest:mutableRequest];
+  [self.dataTask resume];
+}
+
+- (void)stopLoading {
+  [self.dataTask cancel];
+}
+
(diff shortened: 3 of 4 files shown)


MaikuB:
Thanks for the PR. Before I go take a look into in more depth, do you have instructions I could follow to test this PR and ensure the functionality continues to work in the future?

jrevillard:
Hi @MaikuB, here's how to test this PR:

### Prerequisites

You need an OIDC provider with a **self-signed HTTPS certificate**. The easiest way is a local Keycloak:

```bash
# Generate a self-signed cert
openssl req -x509 -newkey rsa:2048 -keyout key.pem -out cert.pem -days 365 -nodes -subj "/CN=localhost"

# Run Keycloak with the self-signed cert
docker run -d --name keycloak-test -p 8443:8443 \
  -v $(pwd)/cert.pem:/etc/keycloak/cert.pem \
  -v $(pwd)/key.pem:/etc/keycloak/key.pem \
  -e KC_BOOTSTRAP_ADMIN_USERNAME=admin \
  -e KC_BOOTSTRAP_ADMIN_PASSWORD=admin \
  quay.io/keycloak/keycloak:26.6 \
  start \
  --hostname=localhost \
  --hostname-strict=false \
  --https-certificate-file=/etc/keycloak/cert.pem \
  --https-certificate-key-file=/etc/keycloak/key.pem
```

Wait for it to start (~15s), then create a client and a test user via the admin console at `https://localhost:8443` (admin/admin):

- Create a client `test-app` (public client, with redirect URI matching the example app, enable direct access grants)
- Create a user with a non-temporary password

Alternatively, via the API:

```bash
# Get admin token
ADMIN_TOKEN=$(curl -sk -X POST "https://localhost:8443/realms/master/protocol/openid-connect/token" \
  -d "client_id=admin-cli" -d "username=admin" -d "password=admin" -d "grant_type=password" \
  | python3 -c "import sys,json; print(json.load(sys.stdin)['access_token'])")

# Create client
curl -sk -X POST "https://localhost:8443/admin/realms/master/clients" \
  -H "Authorization: Bearer $ADMIN_TOKEN" -H "Content-Type: application/json" \
  -d '{"clientId":"test-app","enabled":true,"publicClient":true,"redirectUris":["com.example.app://callback","com.example.app://logout"],"standardFlowEnabled":true,"directAccessGrantsEnabled":true}'

# Create user
curl -sk -X POST "https://localhost:8443/admin/realms/master/users" \
  -H "Authorization: Bearer $ADMIN_TOKEN" -H "Content-Type: application/json" \
  -d '{"username":"testuser","email":"testuser@example.com","emailVerified":true,"firstName":"Test","lastName":"User","enabled":true,"credentials":[{"type":"password","value":"testpass123","temporary":false}]}'
```

### Test with the example app

1. Point the example app (`flutter_appauth/example/`) at your local Keycloak:
   - `discoveryUrl`: `https://localhost:8443/realms/master/.well-known/openid-configuration`
   - `clientId`: `test-app`
   - `redirectUrl`: your redirect URI
   - **`allowInsecureConnections: true`**

2. Run on Android and/or iOS and exercise these flows:
   - **Authorization + token exchange**
   - **Token refresh**
   - **End session / logout**

### What to verify

| Scenario | Expected |
|---|---|
| `allowInsecureConnections: true` against self-signed HTTPS | All flows succeed (discovery, auth, token refresh, end session) |
| `allowInsecureConnections: false` (default) against self-signed HTTPS | All flows fail with a network/SSL error — bypass is NOT active by default |
| `allowInsecureConnections: false` against a valid CA-signed server | All flows succeed normally — no regression |

The key thing to confirm is that the flag is **opt-in only**: when `false` (the default), behavior is identical to the current release.

I've tested this exact setup — Keycloak 26.6 starts correctly with the self-signed cert, the OIDC discovery endpoint responds, and token grants work against it.

jrevillard:
Any update @MaikuB ?

MaikuB:
I still need to get around to testing it out. I was part way through setting up Keycloak and had other priorities to deal with

MaikuB:
I've not gotten to try on iOS yet but ran into the follow problems after following the steps you had with the only differences being that redirect and post logout redirect URLs are set to match what current example app uses

- invalid scopes error is returned when I trigger the "Sign in with auto code exchange" scenario. Note it didn't lead me to a point where I could enter in a username and password that I assume should have happened
- the "Sign in with no code exchange" scenario doesn't work as it looks like even though I have replaced localhost with 10.0.2.2, it still seems to try to open a browser that goes to localhost

Any advice here?

jrevillard:
Sorry for the delayed response @MaikuB. Both issues are in the example app config, not the PR.

## 1. Invalid scopes

The example app includes `api` in scopes — this is a Duende demo server scope, **not a Keycloak scope**. Keycloak rejects the authorization request immediately. Remove it:

```dart
final List<String> _scopes = <String>["openid", "profile", "email"];
```

## 2. localhost vs 10.0.2.2

The two sign-in buttons use **different config sources**:

- **"Sign in with auto code exchange"** → `serviceConfiguration: _serviceConfiguration` (hardcoded to `demo.duendesoftware.com`). Changing `_discoveryUrl`/`_issuer` has no effect. You must update `_serviceConfiguration`:
  ```dart
  final AuthorizationServiceConfiguration _serviceConfiguration =
      const AuthorizationServiceConfiguration(
    authorizationEndpoint: "https://10.0.2.2:8443/realms/master/protocol/openid-connect/auth",
    tokenEndpoint: "https://10.0.2.2:8443/realms/master/protocol/openid-connect/token",
    endSessionEndpoint: "https://10.0.2.2:8443/realms/master/protocol/openid-connect/logout",
  );
  ```

- **"Sign in with no code exchange"** → `discoveryUrl: _discoveryUrl`. Even after updating it to `10.0.2.2`, Keycloak's discovery document returns endpoints with its `--hostname` value. If started with `--hostname=localhost`, discovered endpoints still point to `localhost`. Fix: start Keycloak with `--hostname=10.0.2.2`.

## 3. allowInsecureConnections

Add `allowInsecureConnections: true` to each request — that's the feature this PR adds:

```dart
AuthorizationTokenRequest(
  _clientId, _redirectUrl,
  serviceConfiguration: _serviceConfiguration,
  scopes: _scopes,
  allowInsecureConnections: true,
)
```

## Keycloak startup with correct hostname

```bash
docker run -d --name keycloak-test -p 8443:8443 \
  -v $(pwd)/cert.pem:/etc/keycloak/cert.pem \
  -v $(pwd)/key.pem:/etc/keycloak/key.pem \
  -e KC_BOOTSTRAP_ADMIN_USERNAME=admin \
  -e KC_BOOTSTRAP_ADMIN_PASSWORD=admin \
  quay.io/keycloak/keycloak:26.6 \
  start \
  --hostname=10.0.2.2 \
  --https-certificate-file=/etc/keycloak/cert.pem \
  --https-certificate-key-file=/etc/keycloak/key.pem
```

MaikuB:
Thanks for the info. I'll give it a go again will reach other if I need more help

MaikuB:
Unfortunately, I still cannot test this properly on Android though iOS and macOS work fine. Can you please provide end to end instructions that would allow the example app in this repo to connect to the Keycloak container via an Android emulator? Alternatively this could be instructions on getting the Keycloak container running but use another fork/branch of yours where it has the changes in this PR and updates required in the example app. In both cases, this should be something that has been tested to work end to end

MaikuB:
Other thing I forgot to mention is originally `allowInsecureConnections` was for HTTP. The testing you have mentioned and instructions shared involve HTTPs via a self-signed certification. Would need to be able to validate this PR still works with HTTP

Issue #651, opened 2026-04-29
[FeatureRequest]: Support proxy redirect URLs for providers that require HTTPS redirect URLs


luis901101:
## Summary

Add a `proxyRedirectUrl` parameter to authorization and token exchange requests so that apps can work with OAuth providers that only allow HTTPS redirect URIs.

## Background

Native mobile apps traditionally use custom-scheme deep links (e.g. `com.example.myapp://oauth2redirect`) as redirect URIs. These are registered with the OS so that the browser hands the OAuth callback back to the app. However, some identity providers enforce a strict policy that only HTTPS redirect URIs are permitted. As a result, apps using custom schemes are rejected at the authorization request stage.

At first glance, Universal Links (iOS) and App Links (Android) appear to be a natural solution since they use HTTPS URLs that the OS can route to the app. However, they are not a reliable alternative in this context:

- On iOS and macOS, `ASWebAuthenticationSession` — the mechanism AppAuth uses to present the authorization flow — only terminates and returns a result when the browser navigates to a URL matching a registered **custom scheme**. Its `callbackURLScheme` parameter does not accept `https`, so a Universal Link used as the redirect URI will not be intercepted by the session; the flow stalls or opens in the browser instead.
- On Android, App Links require hosting a `assetlinks.json` file and passing an OS-level domain verification process that can fail silently, be delayed, or be bypassed by the user choosing to open the link in a browser. AppAuth's `RedirectUriReceiverActivity` is also designed around custom-scheme intent filters rather than App Links.

A well-known workaround is to host a lightweight HTTPS endpoint that acts as a proxy: the provider redirects to the HTTPS URL, which immediately issues a redirect to the app's custom scheme. The OAuth flow then completes as normal.

## Problem

Currently the plugin only accepts a single `redirectUrl`. There is no way to tell the plugin to send a different URL to the provider while still intercepting the callback on-device via a custom scheme. Developers who need to integrate with strict providers (e.g. certain enterprise identity providers, government OAuth servers, or providers with hardcoded HTTPS-only validation) cannot use the plugin without a workaround outside of AppAuth's control.

## Proposed solution

Add an optional `proxyRedirectUrl` parameter to `AuthorizationRequest`, `AuthorizationTokenRequest`, and `TokenRequest`:

- When `proxyRedirectUrl` is set, it is sent to the identity provider as the `redirect_uri` parameter in the authorization request, satisfying the HTTPS requirement.
[...]

PR #652, opened 2026-04-29
[flutter_appauth][flutter_appauth_platform_interface] Add support for proxy redirect URLs

luis901101:
## What this PR does

Adds an optional `proxyRedirectUrl` parameter to `AuthorizationRequest`, `AuthorizationTokenRequest`, and `TokenRequest`. This allows apps to work with identity providers that only permit HTTPS redirect URIs by decoupling the URL sent to the provider from the custom-scheme URL used to intercept the callback on-device.

When `proxyRedirectUrl` is set:
- It is sent to the identity provider as the `redirect_uri` in the authorization request.
- `redirectUrl` (the app's custom-scheme deep link) is still used by AppAuth to intercept the browser callback.
- `proxyRedirectUrl` is also sent as `redirect_uri` during token exchange so it matches what was used in the authorization request, as required by OAuth 2.0.

Notes:
The proxy responsible for the final redirect must be configured accordingly, but this is out of the scope of this PR.

When `proxyRedirectUrl` is not set, behaviour is identical to before.

### DIFF
--- a/flutter_appauth/README.md
+++ b/flutter_appauth/README.md
@@ -11,6 +11,7 @@
   - [End session](#end-session)
   - [Handling errors](#handling-errors)
   - [Ephemeral Sessions (iOS and macOS only)](#ephemeral-sessions-ios-and-macos-only)
+  - [Using a proxy redirect URL](#using-a-proxy-redirect-url)
 - [Android setup](#android-setup)
 - [iOS/macOS setup](#iosmacos-setup)
 - [API docs](#api-docs)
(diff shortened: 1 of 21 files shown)


luis901101:
This PR fixes https://github.com/MaikuB/flutter_appauth/issues/651

MaikuB:
Thanks for the PR and apologies for the delay on coming to this. At the moment, I'm going to hold off on this for the following reasons

- This looks like a workaround that strays further away from exposing more on what is officially available with the AppAuth SDKs, which is the primary purpose of the plugin. Note I understand that there's a use case behind the PR and is arguably where a fork is suitable
- Looks like this PR may have some overlap with [this PR](https://github.com/MaikuB/flutter_appauth/pull/662) since both deal with HTTPs-based redirect URLs
- I do not have way to test the changes
- I'd like to see if the AppAuth SDKs can tackle some of the changes. Looks like there is already a related PR at https://github.com/openid/AppAuth-iOS/pull/938

PR #653, opened 2026-05-05
Use class instead of param for state in authorization and end-session requests 

### LINKED ISSUE #646: Allow explicitly null state in EndSessionRequest

pcba-dev:
Fixes https://github.com/MaikuB/flutter_appauth/issues/646 based on https://github.com/MaikuB/flutter_appauth/pull/647#pullrequestreview-4215730072


MaikuB (changes requested):
Thanks for the updated PR. Besides the linter issue, there are a couple of issues I picked up with the PR so far

- [This file](https://github.com/MaikuB/flutter_appauth/blob/master/flutter_appauth/lib/flutter_appauth.dart) needs to be updated to expose the new classes. This is necessary to make it so that developers only need to import a reference the main plugin access the new classes
- Once done, can you test all of the possible permutations on all platforms? I did a test and saw an exception when triggering an end session request with null state (i.e. used the `SuppressedState` class). This occurred on iOS and looks like further changes would be needed

PR #660, opened 2026-06-05
[flutter_appauth] fix:missing a dependency on FlutterFramework

JarvanMo:
Got the following warnings while running:

```
Plugin flutter_appauth has a Package.swift for ios but is missing a dependency on FlutterFramework. Add the following to your Package.swift dependencies:
    .package(name: "FlutterFramework", path: "../FlutterFramework")
And add FlutterFramework as a target dependency:
    .product(name: "FlutterFramework", package: "FlutterFramework")
See https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-plugin-authors for more information.
Plugin flutter_appauth has a Package.swift for macos but is missing a dependency on FlutterFramework. Add the following to your Package.swift dependencies:
    .package(name: "FlutterFramework", path: "../FlutterFramework")
And add FlutterFramework as a target dependency:
    .product(name: "FlutterFramework", package: "FlutterFramework")
```

### DIFF
--- a/flutter_appauth/ios/flutter_appauth/Package.swift
+++ b/flutter_appauth/ios/flutter_appauth/Package.swift
@@ -11,11 +11,17 @@ let package = Package(
     products: [
         .library(name: "flutter-appauth", targets: ["flutter_appauth"])
     ],
-    dependencies: [.package(url: "https://github.com/openid/AppAuth-iOS", exact: "2.0.0")],
+    dependencies: [
+        .package(url: "https://github.com/openid/AppAuth-iOS", exact: "2.0.0"),
+        .package(name: "FlutterFramework", path: "../FlutterFramework")
+    ],
     targets: [
         .target(
             name: "flutter_appauth",
-            dependencies: [.product(name: "AppAuth", package: "AppAuth-iOS")],
+            dependencies: [
+                .product(name: "AppAuth", package: "AppAuth-iOS"),
+                .package(name: "FlutterFramework", path: "../FlutterFramework")
+            ],
             resources: [
                 .process("PrivacyInfo.xcprivacy")
             ],
--- a/flutter_appauth/ios/flutter_appauth/Package.swift
+++ b/flutter_appauth/ios/flutter_appauth/Package.swift
@@ -20,7 +20,7 @@ let package = Package(
             name: "flutter_appauth",
             dependencies: [
                 .product(name: "AppAuth", package: "AppAuth-iOS"),
-                .package(name: "FlutterFramework", path: "../FlutterFramework")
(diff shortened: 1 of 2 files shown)


flutter_appauth/ios/flutter_appauth/Package.swift
```diff
@@ -11,11 +11,17 @@ let package = Package(
...
+        .package(url: "https://github.com/openid/AppAuth-iOS", exact: "2.0.0"),
+        .package(name: "FlutterFramework", path: "../FlutterFramework")
+    ],
     targets: [
         .target(
             name: "flutter_appauth",
-            dependencies: [.product(name: "AppAuth", package: "AppAuth-iOS")],
+            dependencies: [
+                .product(name: "AppAuth", package: "AppAuth-iOS"),
+                .package(name: "FlutterFramework", path: "../FlutterFramework")
```
MaikuB:
```suggestion
                .package(name: "FlutterFramework", package: "FlutterFramework")
```

Looking at step 4 of [this section](https://docs.flutter.dev/packages-and-plugins/swift-package-manager/for-plugin-authors#how-to-add-swift-package-manager-support-to-an-existing-flutter-plugin), what you had before was incorrect

MaikuB (changes requested):
Thanks for the PR. The Flutter team updated instructions with the recent stable release. Will have to ensure this is backwards compatible to see if a bump to the minimum Flutter version is needed. Have you gotten to check this?

 I left a comment where there appears to be a mistake when comparing what the instructions but additionally, could your PR also make similar changes on the macOS side?

JarvanMo:
Updated.

flutter_appauth/ios/flutter_appauth/Package.swift
```diff
@@ -11,11 +11,17 @@ let package = Package(
...
+        .package(url: "https://github.com/openid/AppAuth-iOS", exact: "2.0.0"),
+        .package(name: "FlutterFramework", path: "../FlutterFramework")
+    ],
     targets: [
         .target(
             name: "flutter_appauth",
-            dependencies: [.product(name: "AppAuth", package: "AppAuth-iOS")],
+            dependencies: [
+                .product(name: "AppAuth", package: "AppAuth-iOS"),
+                .package(name: "FlutterFramework", package: "FlutterFramework")
```
MaikuB:
```suggestion
                .product(name: "FlutterFramework", package: "FlutterFramework")
```

Checks failed as looks like another fix is needed

MaikuB (changes requested):
Thanks for making changes but checks failed as there was another issue to address

MaikuB (approved):

PR #662, opened 2026-06-23
[flutter_appauth][iOS] Support HTTPS redirect URIs on iOS 17.4+

KhaleelSH:
## Summary

This adds support for HTTPS redirect URIs on iOS 17.4 and newer by using `ASWebAuthenticationSessionCallback` with `callbackWithHTTPSHost:path:` when the request redirect URL uses the `https` scheme.

The change applies to both default and ephemeral `ASWebAuthenticationSession` flows. Custom-scheme redirects continue to use the existing `callbackURLScheme:` path, and `SFSafariViewController` behavior is unchanged.

## What Changed

- Route default and ephemeral iOS `ASWebAuthenticationSession` flows through the custom iOS external user agent.
- Pass the request redirect URL into the user agent so HTTPS redirect URLs can use the iOS 17.4+ callback API.
- Keep custom-scheme redirects on the legacy callback scheme initializer.
- Guard nil or malformed redirect URLs, including end-session requests without `postLogoutRedirectUrl`, so the HTTPS callback API is only used when a valid HTTPS host is available.
- Treat host-only HTTPS redirect URLs as the root path when constructing the callback.
- Document the required `webcredentials:<host>` Associated Domains setup and matching `apple-app-site-association` configuration.

## Notes

Apple's HTTPS callback API uses associated web credentials domains. It is not the normal `applinks` universal-link flow, so the README calls out the required `webcredentials` entitlement.

HTTPS redirect URIs only work on iOS 17.4 and newer because the required `ASWebAuthenticationSession` callback initializer is unavailable on earlier iOS versions. Apps that still support older iOS versions should keep a custom-scheme redirect fallback registered with their identity provider.

## Testing

- `melos run analyze`
- `melos run test:unit --no-select`
- `melos run build:example_ios`
- `xcrun clang-format --dry-run --Werror flutter_appauth/ios/flutter_appauth/Sources/flutter_appauth/OIDExternalUserAgentIOSNoSSO.m flutter_appauth/ios/flutter_appauth/Sources/flutter_appauth/OIDExternalUserAgentIOSNoSSO.h flutter_appauth/ios/flutter_appauth/Sources/flutter_appauth/AppAuthIOSAuthorization.m`
- `git diff --cached --check`

### DIFF
--- a/flutter_appauth/CHANGELOG.md
+++ b/flutter_appauth/CHANGELOG.md
@@ -1,3 +1,7 @@
+## [12.1.0]
+
+* [iOS] Added support for `https` redirect URIs on iOS 17.4 and newer. When the redirect URL uses the `https` scheme, the plugin now starts the `ASWebAuthenticationSession` using the `callbackWithHTTPSHost:path:` API. This applies to both the default and ephemeral `ASWebAuthenticationSession` external user agents. No Dart API changes are required — pass an `https` redirect URL as usual. Note that the app must declare the matching Associated Domains entitlement (e.g. `webcredentials:<host>`) for the callback to be delivered.
+
 ## [12.0.1]
 
 * Removed assertion around tied to `idTokenHint` and `postLogoutRedirectUrl` parameters passed to the `EndSessionRequest` constructor. This was done as both are optional according to the [OIDC RP-initiated logout specification](https://openid.net/specs/openid-connect-rpinitiated-1_0.html)
--- a/flutter_appauth/README.md
+++ b/flutter_appauth/README.md
@@ -13,6 +13,7 @@
   - [Ephemeral Sessions (iOS and macOS only)](#ephemeral-sessions-ios-and-macos-only)
 - [Android setup](#android-setup)
 - [iOS/macOS setup](#iosmacos-setup)
+  - [Using https redirect URIs on iOS 17.4+](#using-https-redirect-uris-on-ios-174)
 - [API docs](#api-docs)
 - [FAQs](#faqs)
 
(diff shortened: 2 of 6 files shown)


MaikuB:
Thanks for the PR. I left a comment [here](https://github.com/MaikuB/flutter_appauth/pull/652#issuecomment-5155022824) that is relevant to this PR too. For convenience, what I mean to say is I'll be holding off on this one right now. It would be better if the AppAuth iOS SDK handles it as this PR would become obsolete if [this](https://github.com/openid/AppAuth-iOS/pull/938) is merged in and released

KhaleelSH:
That would be the ideal solution. I'm just a bit afraid that that PR is still open since a while, and would take more time to get merged. But anyways, thank you for letting me know you're holding off on this, I believe I'll use my fork until the mentioned PR is merged. Have a nice day.

travishaagen:
@MaikuB I would like to advocate for this PR to be reviewed and merged. I stumbled on this PR, because I also need it.

It's not clear that the maintainers of AppAuth-iOS have even acknowledged https://github.com/openid/AppAuth-iOS/pull/938 and leading authentication providers ([example](https://auth0.com/docs/secure/security-guidance/measures-against-app-impersonation)) have been saying for a while that custom URL schemes are not secure.

MaikuB:
@travishaagen understood however I'm busy ATM as I'm in the process of moving. If you haven't done so already or weren't aware, note that you can create a fork based on this PR and use it directly in your app in the meantime

KhaleelSH:
This branch was updated and conflicts with main branch were fixed.

PR #663, opened 2026-06-24
[flutter_appauth_web] add web support via the AppAuth-JS SDK

eirikb:
Hi @MaikuB

Another go at web support, this time wrapping the official AppAuth-JS (`@openid/appauth`) like you suggested in #70. The previous attempt (#620) was the hand-rolled one you rightly closed.

### What it is

A separate, endorsed federated package, `flutter_appauth_web`, so it gets pulled in automatically on web and you can endorse it via `default_package` (the url_launcher way). It does the authorization code + PKCE flow by delegating to AppAuth-JS, no custom OAuth logic of my own.

The AppAuth-JS bundle is vendored (committed) and built reproducibly from `js/` (pinned `@openid/appauth@1.3.2`, lockfile, one `js/build.sh`), with an Apache-2.0 NOTICE. It's injected at runtime, so app devs don't have to touch their `web/index.html`.

There were a couple of AppAuth-JS quirks I had to wrap around (the redirect handler reads the URL fragment but the code flow returns the response in the query string, and it throws away token error bodies). Those are documented inline right where they happen.

### Honesty bit

I should be upfront: a lot of this was written with AI (Opus 4.8). Buuut I didn't just trust it blindly. I verified it end to end, real authorization-code-with-PKCE round-trips in a headless browser against the live Duende demo IdP and against our own provider (Azure AD B2C federated), token endpoint returning 200 and the URL cleaned up afterwards. And I ran the whole thing through several postmortem/review passes with a few different models, which caught a fair number of bugs across the iterations before I was happy. So AI-written, but human-checked and actually tested, not vibes.

There's also a small `test/` covering the pure URL logic, and it passes `flutter analyze` under strict mode.

### One open thing

There's a bit of repetition in `authorize`/`authorizeAndExchangeCode` when resolving the config. I could dedupe it cleanly if `flutter_appauth_platform_interface` exported `AcceptedAuthorizationServiceConfigurationDetails`, right now reaching into its `src/` would trip `implementation_imports`. I didn't want to reshape your core types in this PR, but happy to follow up if you'd like that.

Only the redirect flow is supported (no popup), same scope as #620. Let me know what you think 🙂


MaikuB:
Sorry for taking a while to come back to you. Have had to deal with some life things. Appreciate the honesty on AI usage. I'll take a look through it and come back with comments/questions as needed. Haven't gotten to give it a spin yet but will the example app work with the demo identity server on the web like it already does on mobile? Wanted to know if there's more setup I need to do to test this out

eirikb:
Hi there

Yes the example should work out of the box. Just need to set the port to 8080:  
```bash
flutter run -d chrome --web-port=8080
```

Just did a local test and this worked:

```bash
git clone https://github.com/eirikb/flutter_appauth --branch feature/eirikb/web-support-appauth-js
cd flutter_appauth/
gg flutter@3 pub run melos bootstrap
cd flutter_appauth/example/
gg flutter@3 run -d chrome --web-port=8080
```

 Note: I used [gg](https://github.com/eirikb/gg/) here just to handle flutter versions.

MaikuB:
Apologies for the delay. I got around to testing it and one odd thing I noticed is that the scenario in to perform the authorisation request to get a code ends up getting tokens as well. Do you know what is going on there?

It looks like a newer release of the AppAuth JS SDK (1.4.0) is out too. Could you bump it and resolve the conflicts too?

Issue #418, opened 2023-04-12
Question: Support for 'code id_token' as responseMode(response_type) in AuthorizationTokenRequest


Overlord21:
Hello,

I have been using the Flutter AppAuth library for implementing OAuth2/OpenID Connect authentication in my app, and I encountered an issue when trying to use the 'code id_token' response_type.

Currently, it seems that the library defaults to the 'code' response_type. However, in my use case, I need the response_type to be 'code id_token'. When I tried to modify the URL manually and used 'code id_token' in the browser, the authentication flow worked correctly. But with the current implementation of the library, I couldn't find a way to set the response_type as 'code id_token'.

Could you please guide me on how to set the response_type to 'code id_token' using the AuthorizationTokenRequest class in the library? If it's not currently supported, I kindly request you to consider adding this feature, as it would be helpful for developers who need to use different response types in their OAuth2/OpenID Connect implementations.

Here's a snippet of my current implementation:

```
final authorizationTokenRequest = AuthorizationTokenRequest(
  Configuration.auth0ClientId,
  Configuration.authRedirectUri,
  issuer: Configuration.authIssuer,
  clientSecret: Configuration.authClientSecret,
  scopes: scopes2,
  nonce: nonce,
  promptValues: ['login'],
//  responseMode: 'code id_token',
);
final AuthorizationTokenResponse? result =
    await appAuth.authorizeAndExchangeCode(authorizationTokenRequest);
```

I tried setting the responseMode variable to 'code id_token' but it seems to be irrelevant.

Thank you for your time and your work on this library. Looking forward to hearing your suggestions.

Best regards,

MaikuB:
> Could you please guide me on how to set the response_type to 'code id_token' using the AuthorizationTokenRequest class in the library? If it's not currently supported, I kindly request you to consider adding this feature, as it would be helpful for developers who need to use different response types in their OAuth2/OpenID Connect implementations.

This plugin is a library is a wrapper for native SDKs that are maintained by others. Consequently, the functionality available depends on what those SDKs support. This should mean that your issue is occurring because those SDKs don't support it and you can look further to confirm. Given they are open source you (and so is this plugin) you can contribute to them if there are features lacking as well

MaikuB:
Had someone I personally know asked about this before so other thing I should mention that if there's something that can be changed on this plugin's side to support this then please look at submitting a PR

PR #666, opened 2026-07-02
[flutter_appauth] Refactor authorization handling to support multiple response types in OAuth flows

kangatahy:
Updated iOS and macOS implementations to utilize ASWebAuthenticationSession for manual implicit authorization. Added responseTypes parameter to various request and response classes, ensuring compatibility with the latest OpenID Connect specifications. Adjusted Android manifest and Gradle dependencies for improved functionality.

### DIFF
--- a/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppAuthRedirectUriReceiverActivity.java
+++ b/flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppAuthRedirectUriReceiverActivity.java
@@ -1,0 +1,29 @@
+package io.crossingthestreams.flutterappauth;
+
+import android.net.Uri;
+import android.os.Bundle;
+import androidx.appcompat.app.AppCompatActivity;
+import net.openid.appauth.AuthorizationManagementActivity;
+
+/**
+ * Receives OAuth redirect URIs for flutter_appauth.
+ *
+ * <p>Implicit flows ({@code id_token token}, etc.) are handled manually so query and fragment
+ * parameters are both parsed. Authorization code flows are forwarded to AppAuth as usual.
+ */
+public class FlutterAppAuthRedirectUriReceiverActivity extends AppCompatActivity {
+  @Override
+  public void onCreate(Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+
+    Uri uri = getIntent() != null ? getIntent().getData() : null;
+    if (FlutterAppauthPlugin.handleManualImplicitRedirectUri(uri)) {
+      finish();
+      return;
+    }
+
+    startActivity(
+        AuthorizationManagementActivity.createResponseHandlingIntent(this, uri));
+    finish();
+  }
+}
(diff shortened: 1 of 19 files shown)


kangatahy:
This PR addresses #418, i personally tested with an oidc provider and used id_token and token response types, worked fine

flutter_appauth/pubspec.yaml
```diff
@@ -12,7 +12,8 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  flutter_appauth_platform_interface: ^12.0.1
+  flutter_appauth_platform_interface:
```
MaikuB, line 15: Please revert this change. Presumably, you made use of this for local testing. Note that you can do by setting melos on your machine as covered [here](https://github.com/MaikuB/flutter_appauth/blob/master/CONTRIBUTING.md)

flutter_appauth_platform_interface/lib/src/authorization_request.dart
```diff
@@ -21,6 +21,7 @@ class AuthorizationRequest extends CommonRequestDetails
         ExternalUserAgent.asWebAuthenticationSession,
     String? nonce,
     String? responseMode,
+    List<String>? responseTypes,
```
MaikuB, line 24: It's not clear that the default response type is `code`. Can you change it so that there is a default value to indicate this and make responseTypes non-nullable. You should also be able to change this from a `List` to a `Set` to further enforce that the values in response type must be unique

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppauthPlugin.java
```diff
@@ -475,6 +498,97 @@ private void performAuthorization(
...
+
+  private boolean isManualImplicitResponseType(@NonNull String responseType) {
+    return ResponseTypeValues.TOKEN.equals(responseType)
+        || ResponseTypeValues.ID_TOKEN.equals(responseType)
+        || "id_token token".equals(responseType)
+        || "token id_token".equals(responseType);
+  }
+
+  private boolean processManualImplicitRedirectUri(@Nullable Uri uri) {
+    if (pendingOperation == null || !pendingOperation.manualImplicit) {
```
MaikuB, line 516: Rather than doing all of this manual processing to extract the values from the `Uri,` have you checked the AppAuth Android SDK around a method that could have built an `AuthorizationResponse` from a `Uri` that would allow reusing the same logic the plugin extract the relevant values to put in the response map?

flutter_appauth/android/src/main/java/io/crossingthestreams/flutterappauth/FlutterAppAuthRedirectUriReceiverActivity.java
```diff
@@ -0,0 +1,29 @@
...
+ * <p>{@link AuthorizationManagementActivity} is always notified so the browser tab opened during
+ * authorization is removed from the back stack.
+ */
+public class FlutterAppAuthRedirectUriReceiverActivity extends AppCompatActivity {
+  @Override
+  public void onCreate(Bundle savedInstanceState) {
+    super.onCreate(savedInstanceState);
+
+    Uri uri = getIntent() != null ? getIntent().getData() : null;
+    FlutterAppauthPlugin.handleManualImplicitRedirectUri(uri);
```
MaikuB, line 23: I sense a code smell of sorts here as the `boolean` return value is not used. The docs you have would indicate the code should branch based on if it's an implicit flow

MaikuB (changes requested):
I've not gotten to give this a good bash but have noticed a few things. Can you take a look? I'll see if Copilot can also assist as scope of this is fairly large

PR #671, opened 2026-08-06
[flutter_appauth] macOS: also resume the authorization flow from the app lifecycle

cristianmgm7:
On macOS the plugin installs a `kAEGetURL` Apple Event handler so an authorization redirect delivered from outside the app's own browser session can resume the pending flow. In a large class of host apps that handler never runs.

`NSAppleEventManager setEventHandler:` **replaces** any existing handler for the event. AppKit installs its own during `finishLaunching` — the one that drives `application:openURLs:` — and plugin registration happens earlier, during nib loading. So AppKit's registration wins and `handleGetURLEvent:` is silently dead in **any app whose delegate implements `application:openURLs:`**, which is any app with deep links.

The symptom is a login that hangs with no error and no cancellation: the redirect reaches the host app, the plugin never sees it, and the pending `OIDExternalUserAgentSession` is never resumed. It shows up with flows whose redirect legitimately returns through a different browser — an emailed magic link, where the user opens the link from their mail client.

**The change:** also register the plugin with `addApplicationDelegate:` and implement `handleOpenURLs:`, so the resume happens on whichever path actually delivers the URL. The two paths are safe together — the flow is cleared after the first successful resume, and `handleOpenURLs:` returns `YES` only when a pending flow accepted the URL, leaving unrelated deep links to the host app. The macOS header declares `FlutterAppLifecycleDelegate` conformance, which `addApplicationDelegate:` requires.

**Verified** on macOS against a Keycloak realm: without this the flow hangs indefinitely; with it the authorization completes normally. iOS already registers via `addApplicationDelegate:` and is untouched by this change.

### DIFF
--- a/flutter_appauth/ios/flutter_appauth/Sources/flutter_appauth/include/flutter_appauth/FlutterAppauthPlugin.h
+++ b/flutter_appauth/ios/flutter_appauth/Sources/flutter_appauth/include/flutter_appauth/FlutterAppauthPlugin.h
@@ -6,6 +6,13 @@
 #import <Flutter/Flutter.h>
 #endif
 
+#if TARGET_OS_OSX
+@interface FlutterAppauthPlugin
+    : NSObject <FlutterPlugin, FlutterAppLifecycleDelegate>
+
+@end
+#else
 @interface FlutterAppauthPlugin : NSObject <FlutterPlugin>
 
 @end
+#endif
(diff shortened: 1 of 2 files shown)


cristianmgm7:
Closing in favour of a branch based on `master` — this one was branched from the `flutter_appauth-v11.0.0` tag, so it conflicts with a since-landed clang-format reflow of the very lines it touches. Same change, reopened cleanly.

PR #677, opened 2026-09-02
[flutter_appauth] Fix macOS build failure caused by PrivacyInfo.xcprivacy symlink

mmahamed:
Replace the PrivacyInfo.xcprivacy symlink with a regular file to fix the macOS build failure during the CpResource phase.


MaikuB:
Presumably it's straightforward but do you have information on how I could reproduce the build failure?

